PR #19, opened 2021-01-22
feat: use api endpoints

### LINKED ISSUE #16: Use Graasp endpoints

pyphilia:
close #16

Don't hesitate to test with a lot of different collections, to test every resource possible.


src/client/utils/date.js
```diff
@@ -0,0 +1,9 @@
+import { DEFAULT_LOCALE } from '../config/constants';
+
+// eslint-disable-next-line import/prefer-default-export
+export const toDate = (published) => {
```
juancarlosfarah, line 4: Maybe call this `formatDate`?

src/client/config/constants.js
```diff
@@ -7,3 +7,22 @@ export const BLOCK_MATH_INDICATOR = '\\[';
...
+  RESOURCE: 'Resource',
+};
+
+export const MIME_TYPES = {
+  HTML: 'text/html',
+  TEXT: 'text/plain',
+};
+
+export const DEFAULT_LOCALE = 'en';
+export const DEFAULT_MEMBER_NAME = 'Guest';
```
juancarlosfarah, line 23: Is it possible to `translate` this using `i18n`?

src/client/components/collection/Summary.js
```diff
@@ -33,22 +34,25 @@ function Summary({
...
+              title={name}
+            />
           </Card>
         </Grid>
         <Grid item sm={12} md={6}>
           <Typography variant="h1" gutterBottom>
             {name}
           </Typography>
-          <Badges views={views} likes={likes} rating={rating} />
+          <Badges views={views} likes={likes} />
```
juancarlosfarah, line 55: Should we not include `ratings`?

src/client/components/collection/Item.js
```diff
@@ -21,6 +23,7 @@ const useStyles = makeStyles((theme) => ({
   cardDescriptionText: {
     '& p': {
       fontSize: 'large',
+      fontFamily: 'arial',
```
juancarlosfarah, line 26: This should probably be factored out, no? I think you can access it via `theme`.

src/client/components/collection/Comments.js
```diff
@@ -19,15 +19,19 @@ const useStyles = makeStyles((theme) => ({
...
   return (
     <>
       <CommentsHeader />
       <Grid container>
         <Grid item xs={12}>
           <List className={classes.root}>
             {comments.map((comment) => {
-              const { id } = comment;
-              return <Comment key={id} comment={comment} />;
+              const { published } = comment;
```
juancarlosfarah, line 33: Why do we used `published` as the key?

src/client/config/constants.js
juancarlosfarah: I see that you do this in the component.

src/client/components/CollectionProvider.js
```diff
@@ -16,7 +16,7 @@ class CollectionProvider extends React.Component {
 
   state = (() => {
     const { data } = this.props;
-    return { ...data };
+    return data;
```
juancarlosfarah, line 19: So why do we not need this anymore?

src/client/components/collection/Badges.js
```diff
@@ -26,11 +25,9 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-function Badges({ views, likes, rating }) {
+function Badges({ views, likes }) {
```
juancarlosfarah, line 28: Why are ratings not included yet?

juancarlosfarah (changes requested):
Thanks for the PR! Some minor changes and questions here.

src/client/components/collection/Badges.js
pyphilia: André said that since most of the ratings are not set, it was better not displaying them rather than displaying 0-rated collections.

src/client/components/collection/Comments.js
pyphilia: Because there is no id for comments.

PR #20, opened 2021-01-22
feat: docker

pyphilia:
I am not confortable with Docker, I tried but currently I have the following error 

```
> Started on port 3000
/node_modules/quill/dist/quill.js:7661
var elem = document.createElement('div');
           ^

ReferenceError: document is not defined
```

Anyone can use this branch to correct it!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM node:alpine
+
+ENV NODE_ENV production
+ENV HOST 0.0.0.0
+ENV PORT 80
+
+# ENV RAZZLE_CUSTOM_VARIABLE x
+
+# Bundle APP files
+COPY build .
+COPY package.json .
+COPY build/public public
+
+RUN yarn install --production
+
+CMD [ "node", "./server.js" ]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+ENV NODE_ENV production
+ENV HOST 0.0.0.0
+ENV PORT 80
+
+# ENV RAZZLE_CUSTOM_VARIABLE x
+
+# Bundle APP files
+COPY build .
+COPY package.json .
+COPY build/public public
```
juancarlosfarah, line 12: Not sure this statement is needed. Shouldn't it be included in 10?

juancarlosfarah:
@hasanagh, can you have a look at this?

juancarlosfarah:
@pyphilia, note that the issue that you are showing with `document` is because `Quill` is not SSR friendly. Are you sure you're using the latest version of the code? I thought I fixed it in my first SSR commit (you can see it in the `Text` component).

hasanagh:
> @hasanagh, can you have a look at this?

I am checking this

hasanagh:
It should be fine now. It was missing the function call in the dynamic import of quill.

hasanagh (approved):

juancarlosfarah (approved):

PR #21, opened 2021-01-25
feat: display all collections on home

pyphilia:
close #17

PR #24, opened 2021-01-26
23/copy

### LINKED ISSUE #22: Handle styles correctly with SSR
This requires to apply SSR correctly for Material styles  https://material-ui.com/styles/advanced/#next-js

### LINKED ISSUE #23: Copy Item in Graasp

pyphilia:
close #23
close #22 

contains #21


src/client/components/collection/CopyButton.js
```diff
@@ -0,0 +1,139 @@
...
+  }, [isSignedIn]);
+
+  const onClick = async () => {
+    if (!isSignedIn) {
+      setShowLoginModal(true);
+    } else {
+      setShowTreeModal(true);
+    }
+
+    setShowTreeModal(true);
```
pyphilia, line 66: remove this line.

pyphilia:
This PR is ready for review, but you can review https://github.com/graasp/graasp-explore/pull/21/files to review a first part, then this PR will be rebased.

It contains the possibility to copy a space in your own workspace, and provides minimal feedback on success or fail (eg. ILS cannot be copied in another ILS).

src/api/collection.js
```diff
@@ -1,21 +1,48 @@
...
+    body: JSON.stringify(body),
+    headers: {
+      ...DEFAULT_POST.headers,
+      cookie,
+    },
+  });
+
+  // error if the !res.ok
+  // or redirect the request to the login page
+  if (!res.ok || res.url?.includes('/login')) {
```
juancarlosfarah, line 42: Factor out `/login`.

src/client/components/Root.js
```diff
@@ -52,20 +52,27 @@ let theme = createMuiTheme({
...
-    <MuiThemeProvider theme={theme}>
-      <I18nextProvider i18n={i18nConfig}>
-        <App />
-        <ToastContainer />
-      </I18nextProvider>
-    </MuiThemeProvider>
-  </div>
-);
+const Root = ({ classes }) => {
+  useEffect(() => {
```
juancarlosfarah, line 56: Can you comment why we need this effect?

src/client/components/collection/CollectionsGrid.js
```diff
@@ -0,0 +1,61 @@
...
+import { useTranslation } from 'react-i18next';
+import Typography from '@material-ui/core/Typography';
+import Grid from '@material-ui/core/Grid';
+import CollectionCard from './CollectionCard';
+
+function CollectionsGrid({ collections }) {
+  const { t } = useTranslation();
+
+  return !collections.length ? (
+    <div className="Main">
```
juancarlosfarah, line 12: What's this class name?

src/client/components/collection/CopyButton.js
```diff
@@ -0,0 +1,139 @@
...
+  const [showTreeModal, setShowTreeModal] = useState(false);
+  const [isSignedIn, setIsSignedIn] = useState(false);
+  const [copying, setCopying] = useState(false);
+
+  const checkUserIsSignedIn = async () => {
+    const res = await (await fetch(IS_AUTHENTICATED_ROUTE, DEFAULT_GET)).json();
+    setIsSignedIn(res);
+  };
+
+  useEffect(async () => {
```
juancarlosfarah, line 46: I really like this!

src/client/components/collection/CopyButton.js
```diff
@@ -0,0 +1,139 @@
...
+    if (!isSignedIn) {
+      setShowLoginModal(true);
+    } else {
+      setShowTreeModal(true);
+    }
+
+    setShowTreeModal(true);
+  };
+
+  const copy = async ({ id: toSpace }) => {
```
juancarlosfarah, line 69: Do we not want to factor out these calls that go to the backend ?

src/server.js
```diff
@@ -64,34 +89,89 @@ const handleRender = (req, res, data) => {
   }
 };
 
+const handleSpaceVieverRender = (req, res) => {
```
juancarlosfarah, line 92: typo in `Viewer`

src/index.js
```diff
@@ -4,6 +4,7 @@ let app = require('./server').default;
 
 if (module.hot) {
   module.hot.accept('./server', () => {
+    // eslint-disable-next-line no-console
```
juancarlosfarah, line 7: Add an issue to add a logger like the one we have for the API. Then we won't get any of these errors.

src/client/utils/common.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const cloneDeep = (el) => JSON.parse(JSON.stringify(el));
```
juancarlosfarah, line 2: Can't use anything in `lodash`? `JSON.parse` and `JSON.stringify` can throw errors.

src/client/config/helpers.js
```diff
@@ -1,11 +1,67 @@
...
 
-export {
-  // eslint-disable-next-line import/prefer-default-export
-  openInNewTab,
+const openContentInNewTab = (content) => {
+  const newWindow = window.open('', '_blank');
+  newWindow.document.write(content);
 };
+
+const openInNewWindow = ({ url, title }) => {
```
juancarlosfarah, line 15: Is this still being used?

src/client/config/env.js
```diff
@@ -1,2 +1,5 @@
-// eslint-disable-next-line import/prefer-default-export
-export const { REACT_APP_GOOGLE_ANALYTICS_ID } = process.env;
+// export const {
```
juancarlosfarah, line 1: Add an issue to solve adding env files into the front end. Probably requires some Webpack stuff. @hasanagh, any ideas?

src/client/config/constants.js
```diff
@@ -28,3 +28,13 @@ export const MEMBER_TYPES = {
...
   CONTRIBUTOR: 'contributor',
 };
+
+export const LOGIN_POP_UP_WIDTH = 450;
+export const LOGIN_POP_UP_HEIGHT = 450;
+
+export const ROOT_ID = 'ROOT';
+export const TREE_VIEW_HEIGHT = 300;
+export const TREE_VIEW_MIN_WIDTH = 350;
+
```
juancarlosfarah, line 38: For the following you can have multiple errors and success codes. Try to use ranges or `response.ok`.

juancarlosfarah (changes requested):
Very nice! I love this PR! :-)

Just some minor questions and comments.

As an aside, try to add comments in the tricky bits. To feature-proof.

Dockerfile
```diff
@@ -1,16 +1,20 @@
...
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+# install app dependencies
+COPY package.json /usr/src/app/
+RUN yarn install
+
+# bundle app source
+COPY . /usr/src/app
+
```
hasanagh, line 14: I dont think we need to specify the full absolute path after defining the Workdir step. you can just use `.`

hasanagh (changes requested):
Thanks for the PR! I had one comment in addition to JC's

src/client/components/Root.js
hasanagh: Should this effect run on each render? cant have an empty dependency list?

src/client/components/collection/CopyButton.js
hasanagh: I think and in order to use an async call inside you effect. you should be doing something like this:
`useEffect(() => {`
  `  (async function checkSignIn() {`
      `await checkUserIsSignedIn();`
  `  })();`
`  }, []);`

check this: https://github.com/facebook/react/issues/14326#issuecomment-441680293

src/client/config/constants.js
pyphilia: These are used to set express response, I don't think I can only set `ok` but if there is a way let me know! I will rename `DEFAULT_SUCCESS_CODE` and forward the status code from the original call.

pyphilia:
I deployed the latest version and apply all the requested changes. I also edited:
- internal requests response. I return a JSON object which is more clean.
- the tree modal switches to login modal when the user get signed out


Side note: the copy actually copy ILS in ILS successfully, but André is working on this.

src/api/authentication.js
```diff
@@ -0,0 +1,22 @@
+import fetch from 'node-fetch';
+import { DEFAULT_ERROR_CODE } from '../client/config/constants';
+import { DEFAULT_GET, formatCookies } from './common';
+import { IS_AUTHENTICATED_ENDPOINT } from './endpoints';
+
+// eslint-disable-next-line import/prefer-default-export
+export const isAuthenticated = async (cookies) => {
```
juancarlosfarah, line 7: Maybe a short comment explaining what this is doing.

src/api/navigation.js
```diff
@@ -0,0 +1,22 @@
+import fetch from 'node-fetch';
+import { DEFAULT_ERROR_CODE } from '../client/config/constants';
+import { DEFAULT_GET, formatCookies } from './common';
+import { GET_NAV_TREE_ENDPOINT } from './endpoints';
+
+// eslint-disable-next-line import/prefer-default-export
+export const getNavTree = async (cookies) => {
```
juancarlosfarah, line 7: Maybe just a quick comment saying what this is fetching.

src/server.js
```diff
@@ -65,20 +88,57 @@ const handleRender = (req, res, data) => {
   }
 };
 
+// redirect to viewer mode of the space
+const handleSpaceViewerRender = (req, res) => {
+  const { id } = req.params;
+  if (!ObjectId.isValid(id)) {
```
juancarlosfarah, line 94: If you're throwing an error here, you should catch it with some middleware so that the server doesn't collapse.

juancarlosfarah: Are you doing that?

src/client/config/constants.js
juancarlosfarah: I see. OK. In that case, you should be explicit. I would use something like this:

https://www.npmjs.com/package/http-status-codes

juancarlosfarah (changes requested):
Looks good! Happy you got it to work. Just some minor questions, comments and suggestions here.

src/server.js
pyphilia: No, it just fails. Maybe we can redirect the user to a 404 page. I'll open an issue with this.

PR #37, opened 2021-02-03
31/sns

### LINKED ISSUE #31: enable email and social media buttons
the sharing buttons in the collection should be enabled

pyphilia:
close #31

I used mock data for the sharing texts (I only used the url and/or name of the collection). If you have better sharing texts for the mail sharing & twitter let me know.
Facebook can share links with a simple link or with an API, but for the latest method you need a facebook app id.

### DIFF
--- a/src/client/components/collection/Badges.js
+++ b/src/client/components/collection/Badges.js
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react';
+import _ from 'lodash';
 import { useTranslation } from 'react-i18next';
 import { makeStyles } from '@material-ui/core/styles';
 import Badge from '@material-ui/core/Badge';
@@ -12,6 +13,11 @@ import {
 import PropTypes from 'prop-types';
 import { Grid, IconButton } from '@material-ui/core';
 import { openInNewTab } from '../../config/helpers';
+import {
+  MAIL_BREAK_LINE,
+  TWITTER_MESSAGE_MAX_LENGTH,
+} from '../../config/constants';
+import { removeTagsFromString } from '../../utils/text';
 
 const useStyles = makeStyles((theme) => ({
   badges: {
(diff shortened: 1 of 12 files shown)


juancarlosfarah:
OK. Please remind me to get you the App ID.

hasanagh (approved):

src/client/components/collection/Badges.js
```diff
@@ -25,8 +26,27 @@ const useStyles = makeStyles((theme) => ({
...
+    openInNewTab(`https://twitter.com/intent/tweet?text=${textToShare}`);
+  };
+
+  const shareOnFacebook = () => {
+    const link = pageLocation;
+    openInNewTab(`https://www.facebook.com/sharer/sharer.php?u=${link}`);
+  };
+
+  const subject = name;
+  const message = pageLocation;
```
juancarlosfarah, line 48: Maybe we can have the content coming from the `t` JSON files. Something simple like: "Check out this collection on Graasp." And then maybe appending the collection description, if any.

src/client/components/collection/Badges.js
```diff
@@ -25,8 +26,27 @@ const useStyles = makeStyles((theme) => ({
...
 
-function Badges({ views, likes }) {
+function Badges({ views, likes, name }) {
   const classes = useStyles();
+  const [pageLocation, setPageLocation] = useState(null);
+
+  useEffect(() => {
+    setPageLocation(window?.location.href);
+  });
+
```
juancarlosfarah, line 36: How's it looking on Twitter and FB? Any previews? It would be good to also include the `og` tags so that we get the right preview. @hasanagh and @hagoptaminian  can help as they did it for the other repos.

juancarlosfarah (changes requested):

pyphilia:
Here some preview 

image

image
(where the background screen is your twitter dash)

pyphilia:
I pushed the og:tags (inspired from other repos), let me know if it is correct.
I also updated the text from your reviews. 
Let me know if we want to change the facebook sharer to use the app id.

src/client/components/common/Seo.js
```diff
@@ -0,0 +1,66 @@
...
+    </Helmet>
+  );
+};
+
+Seo.defaultProps = {
+  lang: DEFAULT_LANG,
+};
+
+Seo.propTypes = {
+  //   url: PropTypes.string.isRequired,
```
juancarlosfarah, line 54: Should we remove this?

juancarlosfarah (approved):
Looking great! 🚀 

Just a minor question about a commented line.

PR #39, opened 2021-02-03
fix: clicking on home should go to home

### LINKED ISSUE #38: clicking on logo should go back to home

hasanagh:
fixed clicking on home

fixes #38

### DIFF
--- a/src/client/components/layout/Header.js
+++ b/src/client/components/layout/Header.js
@@ -5,13 +5,18 @@ import { AccountCircle } from '@material-ui/icons';
 import IconButton from '@material-ui/core/IconButton';
 import { makeStyles } from '@material-ui/core/styles';
 import Logo from '../common/Logo';
+import { HOME_ROUTE } from '../../config/routes';
 
 const useStyles = makeStyles((theme) => ({
   root: {
     flexGrow: 1,
   },
-  grow: {
+  link: {
     flexGrow: 1,
+    display: 'flex',
+    alignItems: 'center',
+    textDecoration: 'none',
+    color: '#fff',
   },
   logo: {
     height: '48px',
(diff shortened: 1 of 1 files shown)


juancarlosfarah (approved):

src/client/components/layout/Header.js
```diff
@@ -27,10 +31,12 @@ function Header() {
     <header>
       <AppBar position="static">
         <Toolbar>
-          <Logo className={classes.logo} />
-          <Typography variant="h6" color="inherit" className={classes.grow}>
-            {t('Explore')}
-          </Typography>
+          <a href="/" className={classes.link}>
```
pyphilia, line 34: I don't know the convention, but maybe you can use `Link` from `react-router`.
And use  `HOME_ROUTE` for the link.

pyphilia (approved):
It was indeed necessary to go back to the home page from a collection! 👍

src/client/components/layout/Header.js
hasanagh: I am not sure if the `Link` component works with static routing. When I tried it, I navigated to home but didn't get any collections (so I am not sure if it works with static routing and couldn't find information online)

pyphilia: @juancarlosfarah  @hasanagh Indeed I have the same result. In fact the routing happens, but since the data are not fetched when mounting `Home` there are no collection (collections are only fetched when rendering from the server). 
- I don't really know the conventions, should we check somehow if the render is done from the server or the client to fetch the data from the component? 
- Maybe we can decide by looking at `collections` in the context, fetch the collections if the value is `null` (while fetching no collection would return  `[]`)
- Or just using a `a` is okay.  

But in any case you can change `HOME_ROUTE` in your `a` tag^^ This does work.

hasanagh: Hey @pyphilia, I added `HOME_ROUTE`. As for the convention I am also not aware of it and checking how to deal with that.

hasanagh: @pyphilia  check [this](https://phrase.com/blog/posts/localized-server-side-rendering-with-react/#The_Navigation_Bar)

pyphilia:
@hasanagh will you update this branch with the changes we mentioned last time? (how we fetch data) or will you use another PR? It's up to you whether you merge this branch now or after.

hasanagh:
> @hasanagh will you update this branch with the changes we mentioned last time? (how we fetch data) or will you use another PR? It's up to you whether you merge this branch now or after.

I will merge this branch. and do the update on the other branch.

PR #55, opened 2021-03-22
add react query

hasanagh:


src/server.js
```diff
@@ -151,14 +158,14 @@ const handleNavTreeEndpoint = (req, res) => {
...
-const handleCollectionRender = (req, res) => {
+const handleCollectionRender = async (req, res) => {
   const { id } = req.params;
   if (!ObjectId.isValid(id)) {
     throw new Error(`id '${id}' is not valid`);
   }
-  getCollection(id, (collection) =>
-    handleRender(req, res, { current: collection }),
-  );
+  const queryClient = new QueryClient();
```
pyphilia, line 166: Is it necessary to create a new `QueryClient` object each time you want to access it? Would it be possible to define it globally?

hasanagh: Yes for security reasons. I quote 

> Create a new QueryClient instance inside of your request handler. This ensures that data is not shared between different users and requests.

src/client/components/collection/Collection.js
```diff
@@ -25,31 +24,31 @@ const useStyles = makeStyles((theme) => ({
...
+  } = collection;
 
   const { pictureId } = image;
   const creator = members.find(({ type }) => type === MEMBER_TYPES.OWNER);
   const contributors = members.filter(({ id: mId }) => mId !== creator.id);
-  const imageUrl =
-    buildImageUrl({ id, pictureId, quality: DEFAULT_PICTURE_QUALITY }) ||
-    ITEM_DEFAULT_IMAGE;
+  const imageUrl = loadingCollection
+    ? image.thumbnailUrl
```
pyphilia, line 49: It is better to use `buildImageUrl` instead of `thumbnailUrl` because the fonction can return a higher quality image.

src/client/components/collection/CollectionCard.js
```diff
@@ -149,14 +145,15 @@ export const CollectionCard = ({ collection = {} }) => {
           content: classes.content,
         }}
       />
-      <CardActionArea onClick={handleCollectionClick}>
+      <Link to={`/collections/${id}`}>
```
pyphilia, line 148: Use `buildCollectionRoute`

src/client/components/collection/CollectionCard.js
```diff
@@ -149,14 +145,15 @@ export const CollectionCard = ({ collection = {} }) => {
...
+        {/*<CardActionArea onClick={handleCollectionClick}>*/}
```
pyphilia, line 149: I think this component comes with a highlight when hovering the children, I don't know if we should use this component and `history`'s `push`, or `Link` as you did.

src/client/components/home/Home.js
```diff
@@ -1,27 +1,120 @@
...
 import Seo from '../common/Seo';
+import { useCollections } from '../../utils/collections';
+import Loader from '../common/Loader';
 
 const useStyles = makeStyles(() => ({
   wrapper: {
     padding: '4vw',
   },
+  root: {
+    padding: '2px 4px',
```
pyphilia, line 21: use `theme` for these numbers. You can check how I did in `Comment.js`.

src/client/components/home/Home.js
```diff
@@ -1,27 +1,120 @@
...
+      setQueried(true);
+    }
+  };
   const classes = useStyles();
   return (
     <>
       <Seo title={APP_NAME} description={APP_DESCRIPTION} author={APP_AUTHOR} />
       <div className={classes.wrapper}>
-        <Typography variant="h2">{t('Home')}</Typography>
+        <Typography variant="h3" align="center">
```
pyphilia, line 68: Maybe factor out the search bar (and the results) in another component.

src/client/components/home/Home.js
```diff
@@ -1,27 +1,120 @@
...
+          <IconButton
+            color="primary"
+            className={classes.iconButton}
+            aria-label="directions"
+            type="submit"
+          >
+            <SearchIcon />
+          </IconButton>
+        </Paper>
+        {isLoading ? (
```
pyphilia, line 99: Maybe include the results in `Paper`? It would avoid the massive margin between the search bar and the results vs `Discover` and `Results`.

src/client/utils/collections.js
```diff
@@ -0,0 +1,74 @@
+import { useQuery } from 'react-query';
+import collectionPlaceholderPng from '../resources/icon.png';
+import { getCollection, getCollections } from '../../api/collection';
+
+const collectionQueryConfig = {
+  staleTime: 1000 * 60 * 60, // time until data in cache considered stale if cache not invalidated
+  cacheTime: 1000 * 60 * 60, // time before cache labeled as inactive to be garbage collected
```
pyphilia, line 7: Factor out?

src/client/utils/collections.js
```diff
@@ -0,0 +1,74 @@
...
+const loadingCollection = {
+  name: 'Loading...',
+  description: 'loading...',
+  loadingCollection: true,
+  comments: [],
+  subitems: [],
+  voteScore: 0,
+  views: 0,
+  image: {
+    backgroundUrl: collectionPlaceholderPng,
```
pyphilia, line 20: I wonder if it would not be better to have a standard placeholder instead of the graasp icon? I think the blue icon vs the grey skeleton clash a bit.

src/client/utils/collections.js
```diff
@@ -0,0 +1,74 @@
...
+
+const loadingCollections = Array.from({ length: 10 }, (v, index) => ({
+  id: `loading-collection-${index}`,
+  ...loadingCollection,
+}));
+
+
+function useCollection(collectionId) {
+  const { data } = useQuery({
+    queryKey: ['collections', collectionId], // cache key for collection detailes
```
pyphilia, line 55: Maybe it would be useful to factor out the `collections` key.

pyphilia (changes requested):
Nice work! I wonder if it would be possible to use react-query in compose..
I left you some comments.

src/client/components/collection/Collection.js
hasanagh: Hey Kim, we are still using the `buildImageUrl` this is only for the placeholder image

src/client/utils/collections.js
hasanagh: mmmm What do you think we should put instead?

pyphilia: https://material-ui.com/components/skeleton/#Facebook.js Maybe just a rectangle? There's an example using cards.

src/client/utils/collections.js
```diff
@@ -0,0 +1,78 @@
...
+  ...loadingCollection,
+}));
+
+function useCollection(collectionId) {
+  const { data, isLoading } = useQuery({
+    queryKey: [COLLECTIONS, collectionId], // cache key for collection detailes
+    queryFn: () => getCollection(collectionId).then((datas) => datas),
+    ...collectionQueryConfig,
+  });
+  return { collection: data ?? loadingCollection, isLoading };
```
pyphilia, line 63: I checked a bit the docs, and would the `placeholderData` works? https://react-query.tanstack.com/guides/placeholder-query-data There seems to have a `isPlaceholderData` as well https://react-query.tanstack.com/reference/useQuery

pyphilia: This could more easily allow to return `useQuery` directly so we can have access to all its data depending on our needs (`isFetchedAfterMount`, `isFetching`, `isIdle`, ...)

src/client/components/collection/CollectionCard.js
```diff
@@ -149,14 +153,24 @@ export const CollectionCard = ({ collection = {} }) => {
...
+      <Link to={buildCollectionRoute(id)}>
+        {isLoading?
+          <Skeleton>
+            <CardMedia
+              className={classes.media}
+              image={thumbnailUrl}
+              title={name}
+              component="img"
+            />
+          </Skeleton>:
```
pyphilia, line 165: There's some spacing missing here. Running prettier will probably resolve it.

src/client/components/home/Home.js
```diff
@@ -1,28 +1,89 @@
...
+  },
+  iconButton: {
+    padding: 10,
+  },
+  divider: {
+    height: 28,
+    margin: 4,
+  },
+  typographyMargin: {
+    margin: '10px 0px',
```
pyphilia, line 35: Go with `theme.spacing` for all numbers you used.

src/client/components/home/Home.js
```diff
@@ -1,28 +1,89 @@
...
+  },
 }));
 
 function Home() {
   const { t } = useTranslation();
-  const { collections } = useContext(CollectionContext);
+  const [queried, setQueried] = useState(false);
```
pyphilia, line 41: Wouldn't is be possible to check for `null` (versus `[]` for empty result)?

src/client/components/home/Home.js
```diff
@@ -1,28 +1,89 @@
...
+          <>
+            <Typography variant="h3" className={classes.typographyMargin}>
+              {t('Search Results')}
+            </Typography>
+            {searchResults.length > 0 ? (
+              <CollectionsGrid collections={searchResults} />
+            ) : (
+              <Typography variant="body1" className={classes.typographyMargin}>
+                {t('No results found.')}
+              </Typography>
```
pyphilia, line 79: Maybe factor out some components in render functions. Eg. `queries ? renderResults() : renderNoResult() `

src/client/components/home/Search.js
```diff
@@ -0,0 +1,67 @@
...
+    width: '100%',
+  },
+  iconButton: {
+    padding: 10,
+  },
+  divider: {
+    height: 28,
+    margin: 4,
+  },
+}));
```
pyphilia, line 25: `theme.spacing`

src/client/components/home/Search.js
```diff
@@ -0,0 +1,67 @@
...
+
+function Search({ handleSearch, isLoading }) {
+  const classes = useStyles();
+  return (
+    <Paper component="div" className={classes.root}>
+      <InputBase
+        id="search"
+        disabled={isLoading}
+        style={{ margin: 8 }}
```
pyphilia, line 34: Factor out in `useStyles` . Plus there's a `margin` property, maybe you just need to set it to `dense`.

src/client/components/home/Search.js
```diff
@@ -0,0 +1,67 @@
...
+        placeholder="Search collections using name or author"
```
pyphilia, line 35: Use `t`

src/client/config/constants.js
```diff
@@ -54,3 +54,10 @@ export const PICTURE_QUALITIES = {
 };
 export const DEFAULT_PICTURE_QUALITY = PICTURE_QUALITIES.LARGE;
 export const MAX_COLLECTION_NAME_LENGTH = 100;
+
+// Cache Keys
+export const COLLECTIONS = 'collections';
```
pyphilia, line 59: Sorry to make you change it once again... What do you think of adding something like `KEY` or something related to react-query? I feel like that `COLLECTIONS` could be anything.

pyphilia (changes requested):
Nice! I really like this library...

I am just wondering if we should go with `isLoading` only or if we should try to handle other cases (or at least have an easier access to it which doesn't lead to add a bunch of props). I feel that providing `isLoading` to all components a bit limited. Maybe `isError` is interesting to propagate (or not), `isFetching`, or maybe just use `status`. What do you think?

src/client/components/home/Home.js
hasanagh: Sorry @pyphilia  but I didn't understand your question. and where exactly?

src/client/utils/collections.js
hasanagh: I agree, but I have to experiment more with it first locally. Not sure if this should be used for our same use case. Anyways this can expand depending on our needs as you said.

src/client/components/home/Home.js
pyphilia: Sorry, the comment was considering line 83 with `{isLoading ? <Loader /> : queried ? renderResults() : null}`
Where I think you can simplify with `{isLoading ? <Loader /> :  renderResults()}`
And `renderResults()` would check for `null` (no search was performed), `[]` (no collection matching the results) and `[...]` rendering some results. That why `queried` is not necessary.

src/client/components/home/Home.js
```diff
@@ -1,28 +1,93 @@
...
+  input: {
+    marginLeft: '20px',
+    flex: 6,
+  },
+  iconButton: {
+    padding: 10,
+  },
+  divider: {
+    height: 28,
+    margin: 4,
```
pyphilia, line 32: Don't forget `theme.spacing` for margin, padding, from line 18 to 32.

src/client/components/home/Search.js
```diff
@@ -0,0 +1,72 @@
...
+import SearchIcon from '@material-ui/icons/Search';
+import Paper from '@material-ui/core/Paper';
+import React from 'react';
+import { makeStyles } from '@material-ui/core';
+import PropTypes from 'prop-types';
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    padding: theme.spacing(0.25, 0.5),
+    margin: '100px 0px',
```
pyphilia, line 14: `theme.spacing`

pyphilia (approved):
Thanks Hassan! I noticed some really minor issues below

juancarlosfarah:
@hasanagh @pyphilia, is this ready to go?

PR #71, opened 2021-11-01
feat: add copy button to collections

### LINKED ISSUE #70: Copy items to own items
Currently only folders can be copied. We should allow items to be copied.

louisewang1:
close #70 
I add the copy button to collections and remove the condition for items


src/client/components/collection/Item.js
```diff
@@ -89,7 +88,7 @@ export const Item = ({ item }) => {
       </Collapse>
 
       <CardActions disableSpacing>
-        {type === ITEM_TYPES.FOLDER && <CopyButton id={id} />}
+        {<CopyButton id={id} />}
```
pyphilia, line 91: You don't need `{ }` around components, but only if you add a condition or use a function/variable.

pyphilia (changes requested):
Congrats on your first PR ;)

PR #73, opened 2021-11-03
72/nextjs

pyphilia:
This PR replaces razzle for nextjs. This was mainly due to an impossibility to use @graasp/ui in explore (react-quill being not ssr friendly).

Using nextjs also solves https://github.com/graasp/graasp-explore/issues/35 

The main changes are:
- No more server / client files. Everything is handled in the components directly.
- No more `react-router`. Nextjs works with `src/pages` as endpoints alongside `next/link`.
- Use of `next/image`, which seems to handle responsiveness and lazy loading.
- Loading `@graasp/ui` only on client rendering with `next/dynamic`

I still need to check the deployment with docker and some build settings.


pyphilia:
Wrong source branch.

PR #74, opened 2021-11-03
feat: replace razzle with nextjs

pyphilia:
This PR replaces razzle for nextjs. This was mainly due to an impossibility to use @graasp/ui in explore (react-quill being not ssr friendly).

Using nextjs also solves https://github.com/graasp/graasp-explore/issues/35 

The main changes are:
- No more server / client files. Everything is handled in the components directly.
- No more `react-router`. Nextjs works with `src/pages` as endpoints alongside `next/link`.
- Use of `next/image`, which seems to handle responsiveness and lazy loading.
- Loading `@graasp/ui` only on client rendering with `next/dynamic`

I still need to check the deployment with docker and some build settings.

@louisewang1 I'll meet with you to discuss these changes!


src/pages/_document.js
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import Document, { Html, Head, Main, NextScript } from 'next/document';
+import { ServerStyleSheets } from '@material-ui/core/styles';
+import theme from '../config/theme';
+import { DEFAULT_LANG } from '../config/constants';
+
+export default class MyDocument extends Document {
```
juancarlosfarah, line 7: Can we rename `MyDocument` to something like `GraaspExplorerDocument`?

src/pages/_app.js
```diff
@@ -0,0 +1,71 @@
...
+  release: `${APP_NAME} ${APP_VERSION}`,
+});
+
+// set up google analytics
+if (typeof window !== 'undefined') {
+  ReactGa.initialize(GOOGLE_ANALYTICS_ID);
+  ReactGa.pageview(window.location.href);
+}
+
+export default function MyApp(props) {
```
juancarlosfarah, line 36: Can we rename `MyApp` to something like `GraaspExplorer` or `GraaspExplorerApp`?

src/pages/_document.js
```diff
@@ -0,0 +1,73 @@
...
+  render() {
+    return (
+      <Html lang={DEFAULT_LANG}>
+        <Head>
+          {/* PWA primary color */}
+          <meta name="theme-color" content={theme.palette.primary.main} />
+          <link
+            rel="stylesheet"
+            href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap"
```
juancarlosfarah, line 16: Do we use Roboto?

juancarlosfarah (approved):
@pyphilia. Excellent PR. 🚀 

Things looks very clean and everything makes sense. Glad we "upgraded" to NextJS. Just added a couple comments about class name changes.

pyphilia:
Thanks for the review!

This PR now also contains the fix for copying public items. We need to call the public endpoint since the user does not have memberships over the target item.
ref #70

pyphilia:
For some incompatibility with docker, I removed any use of next/images.

PR #77, opened 2021-11-08
feat: seperate collections from graasper into Graasp Selection

### LINKED ISSUE #76: seperate graasp selection

louisewang1:
close #76 
Need to change the creator id in .env to the id of Graasper in the production database.

### DIFF
--- a/src/components/home/Home.js
+++ b/src/components/home/Home.js
@@ -57,6 +57,10 @@ function Home() {
       ? null
       : [...new Set(collections?.map(({ creator }) => creator).toArray())],
   );
+  const collectionsGraasper = collections.filter(
+    (collection) =>
+      collection.creator == 'e890a236-3d45-4042-8a95-c91120f4db73',
+  );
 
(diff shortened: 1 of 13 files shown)


src/components/home/Home.js
```diff
@@ -57,6 +57,10 @@ function Home() {
       ? null
       : [...new Set(collections?.map(({ creator }) => creator).toArray())],
   );
+  const collectionsGraasper = collections.filter(
+    (collection) =>
+      collection.creator == 'e890a236-3d45-4042-8a95-c91120f4db73',
```
pyphilia, line 62: Could you use an env variable? Like we do for auth_host, etc.

pyphilia (changes requested):

src/components/home/Home.js
louisewang1: Added the creator id for Graasper in .env.local:
GRAASPER_ID=e890a236-3d45-4042-8a95-c91120f4db73

PR #97, opened 2021-12-16
fix: fix item card width

### LINKED ISSUE #96: Fix Card size when filtering
If you add the last line as following in `AllCollection.js`: 
 ```
  mainWrapper: {
    marginLeft: theme.spacing(5),
    marginBottom: theme.spacing(6),
    width: '100%',
  },
```
It should fix the width of the cards. I think the problem was that since no width was defined, it was set depending on the largest element (the title) and divided the size depending on the grid.. anyway it should do the trick!

### LINKED ISSUE #98: Display tags on item summary

### PR DESCRIPTION
**Pull request by louisewang1**
close #96 
fix the problem that item card width varies

close #98
Display user-defined tags on item summary

### DIFF
--- a/src/components/collection/Collection.js
+++ b/src/components/collection/Collection.js
@@ -59,6 +59,7 @@ const Collection = ({ id }) => {
   const imageUrl = DEFAULT_ITEM_IMAGE_PATH;
 
   const parsedDescription = collection?.get('description');
+  const tags = collection?.get('settings').tags || [];
 
   // todo: views and likes don't exist
   const views = collection?.get('views');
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/components/collection/Collection.js [hunk 1]
```diff
@@ -59,6 +59,7 @@ const Collection = ({ id }) => {
   const imageUrl = DEFAULT_ITEM_IMAGE_PATH;
 
   const parsedDescription = collection?.get('description');
+  const tags = collection?.get('settings').tags || [];
```
pyphilia, line 62: `?.tags`

pyphilia (approved):
Perfect! 🐶

PR #105, opened 2021-12-28
feat: add search box & buttons

### LINKED ISSUE #104: Keyword Search Layout
- Refactor the searching box to enable selections
- Plug in new backend

louisewang1:
close #104 
- new layout for search (see pic in next comment)
  - the author button is also enabled now 
- the old local search is also replaced with the new search with backend and directly search within DB

### DIFF
--- a/src/components/collection/Summary.js
+++ b/src/components/collection/Summary.js
@@ -117,8 +117,14 @@ function Summary({
               ))}
             </>
           )}
-          <Typography variant="h6">{t('Tags')}</Typography>
-          {tags?.length && tags?.map((text) => <Chip label={text} />)}
+          {tags?.length && (
+            <>
+              <Typography variant="h6">{t('Tags')}</Typography>
+              {tags?.map((text) => (
+                <Chip label={text} />
+              ))}
+            </>
+          )}
         </Grid>
       </Grid>
     </div>
--- a/src/components/collection/Summary.js
+++ b/src/components/collection/Summary.js
@@ -118,9 +118,11 @@ function Summary({
             </>
           )}
           <Typography variant="h6">{t('Tags')}</Typography>
-          {tags.map((text) => (
-            <Chip label={text} />
-          ))}
+          {tags?.length ? (
+            tags?.map((text) => <Chip label={text} />)
+          ) : (
+            <Typography variant="body1">{t('None')}</Typography>
+          )}
         </Grid>
       </Grid>
     </div>
--- a/src/components/collection/Summary.js
+++ b/src/components/collection/Summary.js
@@ -118,11 +118,7 @@ function Summary({
             </>
           )}
           <Typography variant="h6">{t('Tags')}</Typography>
-          {tags?.length ? (
-            tags?.map((text) => <Chip label={text} />)
-          ) : (
-            <Typography variant="body1">{t('None')}</Typography>
-          )}
+          {tags?.length && tags?.map((text) => <Chip label={text} />)}
         </Grid>
       </Grid>
     </div>
--- a/src/components/home/Home.js
+++ b/src/components/home/Home.js
@@ -1,6 +1,16 @@
-import { makeStyles } from '@material-ui/core';
-import Typography from '@material-ui/core/Typography';
-import React, { useContext, useState } from 'react';
+import {
+  makeStyles,
+  Radio,
+  RadioGroup,
+  FormControl,
+  FormLabel,
+  FormControlLabel,
+  Tooltip,
+  IconButton,
+  Typography,
+} from '@material-ui/core';
+import HelpIcon from '@material-ui/icons/Help';
+import React, { useContext, useState, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { APP_AUTHOR, APP_DESCRIPTION, APP_NAME } from '../../config/constants';
 import { PUBLISHED_TAG_ID, NEXT_PUBLIC_GRAASPER_ID } from '../../config/env';
@@ -36,12 +46,19 @@ const useStyles = makeStyles((theme) => ({
   typographyMargin: {
     margin: theme.spacing(1.5, 0),
   },
+  searchOptions: {
+    marginTop: theme.spacing(-10),
+    marginLeft: theme.spacing(1),
+  },
 }));
 
 function Home() {
   const { t } = useTranslation();
   const classes = useStyles();
   const [searchResults, setSearchResults] = useState(null);
+  const [typing, setTyping] = useState(null);
+  const [range, setRange] = useState('all');
+  const [keywords, setKeywords] = useState(null);
   const { hooks } = useContext(QueryClientContext);
   const {
     data: collections,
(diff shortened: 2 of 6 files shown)


louisewang1:
image

src/components/collection/Summary.js
```diff
@@ -110,13 +110,17 @@ function Summary({
...
           />
           <Typography variant="h6">{t('Category')}</Typography>
-          {categoriesDisplayed?.map((entry) => (
-            <Chip label={t(entry.name)} />
-          ))}
+          {categoriesDisplayed?.size ? (
+            categoriesDisplayed?.map((entry) => <Chip label={t(entry.name)} />)
+          ) : (
+            <Typography variant="body1">{t('None')}</Typography>
+          )}
```
pyphilia, line 117: Do you prefer showing None or not showing the section at all? I prefer the second option but let me know if you think otherwise.

src/components/home/Home.js
```diff
@@ -36,45 +46,69 @@ const useStyles = makeStyles((theme) => ({
...
+    marginLeft: theme.spacing(1),
+  },
 }));
 
 function Home() {
   const { t } = useTranslation();
   const classes = useStyles();
   const [searchResults, setSearchResults] = useState(null);
+  const [typing, setTyping] = useState(null);
+  const [range, setRange] = useState('all');
```
pyphilia, line 60: Use enum (again you will need to create it again)..

src/components/home/Home.js
```diff
@@ -36,45 +46,69 @@ const useStyles = makeStyles((theme) => ({
...
+      placeholderData: PLACEHOLDER_COLLECTIONS,
+      withMemberships: true,
+    },
   );
+  // const { data: members } = hooks.useMembers(
+  //   null,
+  //   isPlaceholderData
+  //     ? null
+  //     : [...new Set(collections?.map(({ creator }) => creator).toArray())],
+  // );
```
pyphilia, line 75: Remove?

src/components/home/Home.js
```diff
@@ -36,45 +46,69 @@ const useStyles = makeStyles((theme) => ({
...
+  //         (collection) =>
+  //           collection.name.toLowerCase().includes(query) ||
+  //           members
+  //             ?.find(({ id }) => collection.creator === id)
+  //             ?.name.toLowerCase()
+  //             .includes(query),
+  //       ),
+  //     );
+  //   }
+  // };
```
pyphilia, line 100: Remove?

src/components/home/Home.js
```diff
@@ -104,7 +138,35 @@ function Home() {
...
+          handleSearch={handleKeywordInput}
+          handleClick={handleClick}
+          isLoading={isLoading}
+        />
+        <FormControl component="fieldset" className={classes.searchOptions}>
+          <FormLabel component="legend">Search Range</FormLabel>
+          <RadioGroup
+            row
+            aria-label="range"
+            name="search-range-radio-button-group"
```
pyphilia, line 151: is the name actually used?

src/components/home/Home.js
```diff
@@ -36,45 +46,69 @@ const useStyles = makeStyles((theme) => ({
...
+    marginTop: theme.spacing(-10),
+    marginLeft: theme.spacing(1),
+  },
 }));
 
 function Home() {
   const { t } = useTranslation();
   const classes = useStyles();
   const [searchResults, setSearchResults] = useState(null);
+  const [typing, setTyping] = useState(null);
```
pyphilia, line 59: Not sure about the typing namenclature. Shouldn't it be searchInput? Or keywords?

src/components/home/Home.js
```diff
@@ -104,7 +138,35 @@ function Home() {
...
           {t('Browse Open Educational Resources')}
         </Typography>
-        <Search handleSearch={handleSearch} isLoading={isLoading} />
+        <Search
+          handleSearch={handleKeywordInput}
+          handleClick={handleClick}
+          isLoading={isLoading}
+        />
+        <FormControl component="fieldset" className={classes.searchOptions}>
+          <FormLabel component="legend">Search Range</FormLabel>
```
pyphilia, line 147: Translate.

src/components/home/Home.js
```diff
@@ -104,7 +138,35 @@ function Home() {
...
+          <RadioGroup
+            row
+            aria-label="range"
```
pyphilia, line 150: I think this needs to be translated.

src/components/home/Home.js
```diff
@@ -104,7 +138,35 @@ function Home() {
...
+            name="search-range-radio-button-group"
+            value={range}
+            onChange={handleRangeChange}
+          >
+            <FormControlLabel value="all" control={<Radio />} label="All" />
+            <FormControlLabel value="title" control={<Radio />} label="Title" />
+            <FormControlLabel value="tag" control={<Radio />} label="Tag" />
```
pyphilia, line 157: Use enum. And translate.

src/components/home/Home.js
```diff
@@ -104,7 +138,35 @@ function Home() {
...
+            <FormControlLabel
+              value="author"
+              control={<Radio />}
+              label="Author"
+            />
+            <Tooltip title="Use | or & for union/intersection of multiple keywords if choose 'Tag' or 'All'">
```
pyphilia, line 163: Not sure we should let the user add such syntax. I think we need to search for some examples, I don't know if a button would be intuitive. We can discuss with Paul (designer) as well. For now I would assume an intersection.

pyphilia (changes requested):
Nice, I look forward this new search bar!
A general comment: 
- You should update the search bar component instead of adding the options in `Home.js` 
For the interface, could you use the primary color (blue) instead of the red/pink color?

src/components/home/Home.js
louisewang1: It says only typescript file could have enum, so I use a dict instead

src/components/home/Home.js
louisewang1: Ok I would assume intersection, and I'll create a new issue for this feature

src/components/collection/Summary.js
```diff
@@ -118,9 +118,7 @@ function Summary({
             </>
           )}
           <Typography variant="h6">{t('Tags')}</Typography>
-          {tags.map((text) => (
-            <Chip label={text} />
-          ))}
+          {tags?.length && tags?.map((text) => <Chip label={text} />)}
```
pyphilia, line 121: Include the title? So we don't only show the title in case there's no tags.

src/components/home/Home.js
```diff
@@ -42,40 +46,47 @@ function Home() {
...
-  const { data: members } = hooks.useMembers(
-    null,
-    isPlaceholderData
-      ? null
-      : [...new Set(collections?.map(({ creator }) => creator).toArray())],
+  const { data: collections, isLoading } = hooks.usePublicItemsWithTag(
+    PUBLISHED_TAG_ID,
+    {
+      placeholderData: PLACEHOLDER_COLLECTIONS,
+      withMemberships: true,
```
pyphilia, line 57: I removed this property. If you need the memberships you should use `useItemMemberships` (or similar)

src/components/home/Home.js
```diff
@@ -42,40 +46,47 @@ function Home() {
...
+  };
+
+  const handleClick = () => {
+    if (searchInput.split(' ').length > 1) {
+      const keywordArray = searchInput
+        .split(' ')
+        .map((keyword) => keyword.trim());
+      setKeywords(keywordArray.join('&'));
+    } else {
+      setKeywords(searchInput);
```
pyphilia, line 82: I might be wrong, but probably you don't even need to check for an array, the array process should work on a string as well. The only difference I see is that it won't send a string anymore, but it doesn't matter, does it?

src/components/home/Search.js
```diff
@@ -26,43 +36,76 @@ const useStyles = makeStyles((theme) => ({
...
-        <SearchIcon />
-      </IconButton>
-    </Paper>
+    <>
+      <Paper component="div" className={classes.root}>
+        <InputBase
+          id="search"
+          disabled={isLoading}
+          className={classes.search}
+          placeholder={t('Search collections using name or author')}
```
pyphilia, line 61: Should we edit the placeholder text?

src/components/home/Search.js
```diff
@@ -26,43 +36,76 @@ const useStyles = makeStyles((theme) => ({
...
+        <FormLabel component="legend">{t('Search Range')}</FormLabel>
+        <RadioGroup row value={range} onChange={handleRangeChange}>
+          {Object.values(RANGES).map((entry) => (
+            <FormControlLabel
+              value={entry.value}
+              control={<Radio color="primary" />}
+              label={t(entry.title)}
+            />
+          ))}
+          <Tooltip title="Use | or & for union/intersection of multiple keywords if choose 'Tag' or 'All'">
```
pyphilia, line 92: Add a todo that we will later assume intersection.

pyphilia (changes requested):
Thanks for the PR! 🐸

src/components/home/Search.js
louisewang1: I also commented out this tooltip to avoid confusion

src/components/home/Home.js
louisewang1: Yes I don't need to check that.
I added a check for range, because in case of title, the input will be considered a single string. The main reason is in case of title, the backend uses `LIKE '%keyword%` , which I think works better than full text search in case of title

pyphilia (approved):

PR #106, opened 2021-12-31
95/cypress

louisewang1:
ref #95 
Add the skeleton and basic tests for explore. Still need refactor and more tests. Need to fix bugs.


cypress/integration/home-page/collections.spec.js
```diff
@@ -0,0 +1,26 @@
...
+
+    cy.get('div + h3').first().should('have.text', 'Graasp Selection');
+    cy.get('h5 + h3').last().should('have.text', 'Discover');
+  });
+
+  it('display collections', () => {
+    // verify 5 item cards are displayed
+    const itemGrids = cy.get(DISCOVER_GRIDS_SELECTOR);
+    itemGrids.children().should('have.length', 5);
+  });
```
pyphilia, line 25: I think you can merge these tests. But I feel you already tested these in another place?

cypress/support/selector.js
```diff
@@ -0,0 +1,27 @@
...
+  'header > div > div > h6:nth-child(3)';
+export const TITLE_TEXT_SELECTOR = 'main > h3:nth-child(3)';
+export const MY_LIST_TITLE_TEXT_SELECTOR = 'main > h3:nth-child(1)';
+export const HEADER_GRAASP_EXPLORER_SELECTOR =
+  'header > div > div > h6:nth-child(2)';
+export const HEADER_MY_LIST_SELECTOR = 'header > div > div > h6:nth-child(4)';
+export const DISCOVER_GRIDS_SELECTOR = '#__next > div > div > div:nth-child(6)';
+export const ITEM_SUMMARY_TITLE_SELECTOR = 'div > div > div > h1';
+export const ITEM_SUMMARY_CHILDREN_GRIDS_SELECTOR =
+  '#__next > div > div > div:nth-child(3) > div:nth-child(2)';
```
pyphilia, line 27: All your selectors will break very easily. You should add `id` or `className` to html tags to have something more customized but more importantly contextualized. Because in the current case, if we add one wrapper everything will break.

cypress/support/server.js
```diff
@@ -0,0 +1,425 @@
...
+        const parentItem = getItemById(items, body.parentId);
+        path = `${parentItem.path}.${path}`;
+      }
+      newItem = { ...item, id: newId, path };
+      items.push(newItem);
+      // todo: do for all children
+      return reply({
+        statusCode: StatusCodes.OK,
+        body: newItem,
+      });
```
pyphilia, line 265: I think you can change this, so it does nothing.

cypress/support/server.js
```diff
@@ -0,0 +1,425 @@
...
+      ),
+    },
+    ({ reply }) => {
+      if (shouldThrowError) {
+        reply({ statusCode: StatusCodes.BAD_REQUEST, body: null });
+        return;
+      }
+      reply([items[0]]);
+    },
+  ).as('getItemCategories');
```
pyphilia, line 424: Two problems here:
- the alias is the same as the one above
- maybe it's interesting to actually find the correct categories given `items`. Otherwise your browse page won't work at all.

package.json
```diff
@@ -30,6 +33,9 @@
     "@testing-library/user-event": "12.6.0",
     "clsx": "1.1.1",
     "cookie-parser": "1.4.5",
+    "cypress": "9.2.0",
+    "cypress-file-upload": "5.0.8",
```
pyphilia, line 37: You can remove this, we will never add an upload component.

src/components/home/LevelCollectionsPage.js
```diff
@@ -1,23 +1,13 @@
...
-  },
-  divider: {
-    marginTop: theme.spacing(2),
-  },
-}));
-
 const LevelCollectionsPage = ({ selectedOptions }) => {
   const { t } = useTranslation();
-  const classes = useStyles();
+  // const classes = useStyles();
```
pyphilia, line 10: Remove?

src/config/routes.js
```diff
@@ -6,10 +6,6 @@ export const buildPerformViewItemRoute = (id = ':id') =>
   `${GRAASP_PERFORM_HOST}/${id}`;
 export const HOME_ROUTE = '/';
 export const ALL_COLLECTION_ROUTE = '/all-collections';
```
pyphilia, line 8: Since this exists there, why do you need to define it again? (supper/constants)

pyphilia (changes requested):
I think most of the necessary setup is there, good job!
Some comments:
- I see you removed the Header and the Footer. I remember working on the layout so that if there's no collections (the page is empty) the footer is still at the bottom of the page (and not in the middle). Though it's not really important, could you tell me what you did exactly to keep the footer and header even though they are not in the components?
- Could you add this file at the same location? It's the automatic build in github. https://github.com/graasp/graasp-compose/blob/master/.github/workflows/cypress.yml And try to use localhost:3005 for REACT_APP_API_HOST (and probably have to change the env variables name in this workflow)

src/config/routes.js
louisewang1: I think it is because of `next.js`, I'm not able to import from `src/config` folder. If I try to do so, it will throw an error saying it could not destructure `getRuntimeConfig` from `getConfig()`. That's why I created the constants again under cypress folder

cypress/integration/home-page/collections.spec.js
louisewang1: There are similar tests for all-collections page, this one is for home page. They are checking similar things, just for different pages.

louisewang1:
> * I see you removed the Header and the Footer. I remember working on the layout so that if there's no collections (the page is empty) the footer is still at the bottom of the page (and not in the middle). Though it's not really important, could you tell me what you did exactly to keep the footer and header even though they are not in the components?

I moved the `<Header>` and `<Footer>` to `<Wrapper>`, to remove some redundancy. Actually previously the layout among the header, side menu, and the main page is a bit messy, and also repeated, so I refactored it. I set the `position` to fixed, so when user scroll down the page, the user could still access the header and navigate to other pages. Locally the header and footer works fine as before. Let me know if you have any concerns/suggestions

louisewang1:
I will rebase it before I merge ;)

cypress/support/commands.js
```diff
@@ -0,0 +1,105 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import 'cypress-localstorage-commands';
```
pyphilia, line 2: Do you need localstorage?

cypress/support/commands.js
```diff
@@ -0,0 +1,105 @@
...
+  'visitAndMockWs',
+  (visitRoute, sampleData, wsClientStub) => {
+    cy.setUpApi(sampleData);
+    cy.visit(visitRoute, {
+      onBeforeLoad: (win) => {
+        cy.stub(win, 'WebSocket', () => wsClientStub);
+      },
+    });
+  },
+);
```
pyphilia, line 83: Remove if you don't use.

pyphilia (approved):

PR #112, opened 2022-01-13
Add flag item button

louisewang1:
close #112 

Added a button for reporting item in the summary page


src/components/collection/Summary.js
```diff
@@ -54,12 +64,30 @@ function Summary({
...
+  }
+
+  const onFlag = () => {
+    postFlagItem({
+      flagId: selectedFlag.id,
+      itemId,
+    });
+    setOpen(false);
+  };
+
```
pyphilia, line 90: Do you think we can create a whole component `FlagItemButton`? And import it from graasp-ui?

pyphilia: Probably the button will also contain the dialog (so both stay together).

pyphilia (changes requested):

src/components/collection/Summary.js
louisewang1: I tried to add an additional component in graasp-ui, however it's not functioning well. When the dialog is opened, no matter where I click, it will also trigger a click on the flagItemButton, (so it will always trigger an 'open dialog' action, when user click anywhere when the dialog is open, even if the mouse-pointer is not on the dialog window). As a result the dialog can't be closed once it's open.
So I created a component for the FlagItemButton in explore instead. I use the same piece of code and it's working well. Please let me know if you have any idea, I'd love to add it to graasp-ui.

pyphilia (approved):

PR #117, opened 2022-01-19
fix: fix test after rebase

### LINKED ISSUE #116: Some tests fail after rebase
Fix tests after rebase

louisewang1:
Some tests not working properly after rebase. This PR is to fix those.

close #116


pyphilia (changes requested):
Don't forget to run tests before merging. But actually could you rename `.github/workflow` to `.github/workflows`? This should enable auto testing on each push, and will avoid this situation to occur again. Don't forget to add MOCK env variables in this file as well if explorer needs some to run.

PR #119, opened 2022-01-20
feat: category multiselect

### LINKED ISSUE #85: Show skeleton when loading items
https://v4.mui.com/components/skeleton/#skeleton

### LINKED ISSUE #94: improve category search with multiselecting

louisewang1:
enable multiselect for category menu
add skeleton when loading the category list
close #94 
close #85

### DIFF
--- a/src/components/home/AllCollections.js
+++ b/src/components/home/AllCollections.js
@@ -20,6 +20,7 @@ import MenuOpenIcon from '@material-ui/icons/MenuOpen';
 import BookmarkIcon from '@material-ui/icons/Bookmark';
 import HighlightOffIcon from '@material-ui/icons/HighlightOff';
 import AddCircleOutlineIcon from '@material-ui/icons/AddCircleOutline';
+import Skeleton from '@material-ui/lab/Skeleton';
 import Seo from '../common/Seo';
 import {
   APP_AUTHOR,
@@ -29,6 +30,7 @@ import {
   GRAASP_BUILDER_URL,
   LEVEL,
   DISCIPLINE,
+  ALL,
 } from '../../config/constants';
 import CollectionsGrid from '../collection/CollectionsGrid';
 import { QueryClientContext } from '../QueryClientContext';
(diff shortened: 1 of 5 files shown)


src/components/home/AllCollections.js
```diff
@@ -304,8 +339,11 @@ function AllCollections() {
               />
             </>
           )}
-          {(selectedOptions[0] || selectedOptions[1]) && (
-            <LevelCollectionsPage selectedOptions={selectedOptions} />
+          {(selectedLevel?.length > 0 || selectedDiscipline?.length > 0) && (
```
pyphilia, line 342: If I'm not mistaken, line 342 is the opposite of line 324. Perhaps you could juste use `condition ? x : y` instead of having both conditions.

pyphilia, line 342: If I'm not mistaken, line 342 is the opposite of line 324. Perhaps you could juste use `condition ? x : y` instead of having both conditions.

src/components/home/LevelCollectionsPage.js
```diff
@@ -5,31 +5,23 @@ import { useTranslation } from 'react-i18next';
...
-    allCategories?.map((entry) => [entry.name, entry.id]),
-  );
-
-  const categories = selectedOptions?.filter(Boolean);
-  const categoryIds = categories?.map((entry) => categoriesMap?.get(entry));
+  // join all selected categories by comma
+  const categoryIds = [
+    selectedDiscipline.join(','),
+    selectedLevel.join(','),
+  ].filter(Boolean);
```
pyphilia, line 16: It's not good to rely on index to provide ids by category types. 
Can't you use a single string instead?

pyphilia (changes requested):
Thanks for the PR. I've added some comments, I'm pretty concerned about how you fetch items by category though, we can discuss if needed.

src/components/home/LevelCollectionsPage.js
louisewang1: The category types doesn't matter here. It's just group the ids belong to the same type together, the order is arbitrary

louisewang1:
Also add keyboard support for the search button click

PR #126, opened 2022-02-17
122/favorite itemsn

### LINKED ISSUE #122: Favorite collections and like/saved items
- favorite list
- liked items
- save items (optional)

louisewang1:
close #122


src/components/collection/Collection.js
```diff
@@ -61,9 +62,9 @@ const Collection = ({ id }) => {
   const parsedDescription = collection?.get('description') || '';
   const tags = collection?.get('settings')?.tags || [];
 
-  // todo: views and likes don't exist
+  // todo: views don't exist
   const views = collection?.get('views');
-  const likes = collection?.get('likes');
+  const likes = parseInt(likeCount?.count, 10);
```
pyphilia, line 67: I don't think it's normal to have to parseInt it in the frontend. Couldn't it have been dealt with in the backend directly? Most probably you don't even want to have an object {count} but only the number?

src/components/collection/Summary.js
```diff
@@ -52,6 +60,7 @@ const useStyles = makeStyles((theme) => ({
   reportButton: {
     display: 'flex',
   },
+  favoriteButton: {},
```
pyphilia, line 63: remove?

src/components/collection/Summary.js
```diff
@@ -77,12 +86,27 @@ function Summary({
...
+  const { mutate: deleteItemLike } = useMutation(
+    MUTATION_KEYS.DELETE_ITEM_LIKE,
+  );
+
   const [open, setOpen] = useState(false);
   const [selectedFlag, setSelectedFlag] = useState(false);
 
   const { data: flags } = hooks.useFlags();
 
+  if (isMemberLoading) return null;
```
pyphilia, line 104: Show loader? I don't think it's a good to display nothing to the user on load.

pyphilia: Or you don't need to check this if you check if the variable exists (as you did below).

src/components/collection/Summary.js
```diff
@@ -91,6 +115,43 @@ function Summary({
     setOpen(false);
   };
 
+  const handleFavorite = () => {
+    updateFavoriteItem({
```
pyphilia, line 119: It's actually editing the member right? Perhaps we can create a whole new mutations in queryclient, which would handle the whole process. For instance `addItemInFavorite` and `removeItemInFavorite` given the member's id and the item id, and these mutations would create the arrays correctly. I think it will nicely simplify the frontend.

pyphilia (changes requested):
Thanks for the PR! I've left some comments!

src/components/collection/Summary.js
louisewang1: I'm not very sure about this one: As we use the endpoint 'edit member' anyways, with new mutation, we still need to pass in the old favoriteItems, and do the operations to the array. It's a bit of trouble to set the 'extra' field in mutation functions, so I would prefer keep it in frontend. Let me know what you think!

src/components/collection/Summary.js
```diff
@@ -91,6 +113,72 @@ function Summary({
...
+        >
+          <img
+            alt="Creative Commons License"
+            className={classes.icon}
+            src="https://i.creativecommons.org/l/by-nc-sa/4.0/88x31.png"
+          />
+        </a>
+      );
+    }
+    return <></>;
```
pyphilia, line 179: Use graasp-ui soon to come icon.

pyphilia (changes requested):
I've left a minor comment, but I'd like to see the result once you've used the component from graasp-ui! And the new mutations I've suggested.

louisewang1:
I also reworked the my-list page so now it's working!

louisewang1:
image

louisewang1:
I made another layout, maybe you can pick one from the two I've uploaded... @pyphilia
Screenshot from 2022-03-02 14-17-24

src/components/collection/Summary.js
```diff
@@ -91,6 +130,39 @@ function Summary({
     setOpen(false);
   };
 
+  const handleFavorite = () => {
+    updateFavoriteItem({
+      memberId: member.get('id'),
+      extra: member?.get('extra'),
+      itemId,
+      action: 'add',
```
pyphilia, line 138: I would have created two mutations instead. Add, and remove, instead of depending on some string.

src/components/home/MyList.js
```diff
@@ -1,92 +1,111 @@
...
+        </TabPanel>
+        <TabPanel value={tab} index={1}>
+          <CollectionsGrid
+            collections={likedCollections}
+            isLoading={isLoading}
+          />
+        </TabPanel>
+        <TabPanel value={tab} index={2}>
+          <CollectionsGrid collections={ownCollections} isLoading={isLoading} />
+        </TabPanel>
```
pyphilia, line 108: I think you should create new components for each tab. This will split a bit the mutations, and maybe if you have time you can check out `React.lazy` https://reactjs.org/docs/code-splitting.html which can improve loading performance. I think it will fit well with these tabs.

src/components/home/TabPanel.js
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+
+function TabPanel({ children, value, index }) {
+  return (
+    <div
+      role="tabpanel"
+      hidden={value !== index}
+      id={`vertical-tabpanel-${index}`}
+      aria-labelledby={`vertical-tab-${index}`}
```
pyphilia, line 11: do you need these?

pyphilia (changes requested):
I would go with the second proposition, though I think it's confusing to have two "headers". We can discuss again with Paul.

Could you also add layout tests? Especially check cases where the user doesn't  have favorites/likes.

src/components/home/MyList.js
louisewang1: I created new components for each tab. For `React.lazy`, I tried with it but it throw an error saying the REACTDOM does not support Suspense at this moment. Maybe I take care of this later.

louisewang1:
I will take care of the tests in next PR ;)

src/components/home/TabPanel.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+TabPanel.propTypes = {
+  children: PropTypes.node,
+  index: PropTypes.number.isRequired,
+  value: PropTypes.number.isRequired,
+};
+
+TabPanel.defaultProps = {
+  children: {},
```
pyphilia, line 24: `children: null` might work.

pyphilia (approved):
⭐

PR #131, opened 2022-03-14
124/download button

louisewang1:
- Add a download button next to copy buttons
- In case of error, a dialog will pop up


src/components/collection/DownloadButton.js
```diff
@@ -0,0 +1,93 @@
...
+  const { t } = useTranslation();
+  const { useMutation } = useContext(QueryClientContext);
+  const {
+    mutate: downloadItem,
+    data,
+    isLoading,
+    isError,
+  } = useMutation(MUTATION_KEYS.DOWNLOAD_ITEM);
+
+  const [dialogOpen, setDialogOpen] = useState(false);
```
pyphilia, line 28: What about a toast instead a dialog (so you don't need to worry about opening/closing as well as the button)? Check react-toastify and `notifier.js` where all trigger happens! Actually you can trigger the toast error directly from query-client -> I think by default you already created a routine, so you can reuse it in `notifier` ^^

pyphilia (changes requested):
Thanks for the PR! I've left a comment about the dialog, which might be more interesting to have a toast instead. We can discuss if you don't know how to proceed! 🌷

PR #134, opened 2022-03-23
133/item metadata

### LINKED ISSUE #133: Add summary page for single item
When click on a single item, should first open up a summary (metadata) page instead of directly play the content

louisewang1:
close #133 

This PR mainly add an extra layer to non-folder item, to display its metadata, instead of open it in player directly.
I change `Item` to `ChildrenCard`, which is similar to `CollectionCard`, while with `Collection`, depends on it's type, it displays children items, or just display a play button to switch to player.
We can further discuss @pyphilia

### DIFF
--- a/src/components/collection/Collection.js
+++ b/src/components/collection/Collection.js
@@ -17,7 +17,7 @@ import {
 import { QueryClientContext } from '../QueryClientContext';
 import { PLACEHOLDER_COLLECTION } from '../../utils/collections';
 import {
-  buildPerformViewItemRoute,
+  buildPlayerViewItemRoute,
   DEFAULT_ITEM_IMAGE_PATH,
   ITEM_TYPES,
 } from '../../config/constants';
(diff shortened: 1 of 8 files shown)


src/components/collection/Collection.js
```diff
@@ -86,8 +104,25 @@ const Collection = ({ id }) => {
           isLoading={isLoading}
         />
         <Divider className={classes.divider} />
-        <Items parentId={id} />
-        <Divider className={classes.divider} />
+        {type === ITEM_TYPES.FOLDER ? (
+          <>
+            <Items parentId={id} />
+            <Divider className={classes.divider} />
+          </>
```
pyphilia, line 111: Would we want a link from the folder as well to see all the children at once in player?

src/components/home/Home.js
```diff
@@ -83,7 +83,7 @@ function Home() {
       (collection) => collection.creator === NEXT_PUBLIC_GRAASPER_ID,
     );
 
-    if (collectionsGraasper.isEmpty()) {
+    if (collectionsGraasper?.isEmpty()) {
```
pyphilia, line 86: If `collectionsGraasper` in `undefined` it won't display null, will it be the correct behavior?

PR #137, opened 2022-03-31
128/UI

louisewang1:
UI re-worked according to the design by @Paul

ref #128


src/components/collection/Summary.js
```diff
@@ -59,10 +59,15 @@ const useStyles = makeStyles((theme) => ({
     justifyContent: 'center',
   },
   card: {
-    width: '100%',
     marginRight: theme.spacing(2),
+    borderRadius: 35,
+    aspectRatio: 1,
+    boxShadow:
+      '0px 0px 2px 2px rgb(0 0 0 / 24%), 0px 0px 4px 6px rgb(0 0 0 / 18%), 0px 0px 8px 10px rgb(0 0 0 / 12%), 0px 0px 16px 16px rgb(0 0 0 / 8%)',
```
pyphilia, line 66: Is it possible to factor out this shadow string? so it will be easier to change everywhere without changing all files.

src/components/collection/Summary.js
```diff
@@ -74,10 +79,16 @@ const useStyles = makeStyles((theme) => ({
...
   },
+  favoriteButton: {
+    marginLeft: theme.spacing(0.5),
+  },
   icon: {
     marginTop: theme.spacing(1),
     borderWidth: 0,
   },
+  description: {
+    minHeight: 100,
```
pyphilia, line 90: Factor out this number?

src/components/layout/Footer.js
```diff
@@ -22,7 +22,7 @@ function Footer() {
 
   return (
     <footer>
-      <AppBar position="fixed" color="primary" className={classes.appBar}>
+      <AppBar position="fixed" color="#fff" className={classes.appBar}>
```
pyphilia, line 25: secondary?

src/components/layout/Header.js
```diff
@@ -36,12 +37,19 @@ const useStyles = makeStyles((theme) => ({
...
+  header: {
+    boxShadow: HEADER_SHADOW_EFFECT,
+  },
   link: {
     textDecoration: 'none',
     color: '#fff',
   },
   logo: {
-    fill: '#fff',
+    fill: '#000',
```
pyphilia, line 51: it's supposed to be black? Factor out?

src/components/layout/Header.js
```diff
@@ -82,9 +90,9 @@ function Header() {
 
   return (
     <>
-      <header>
-        <AppBar position="fixed">
-          <Toolbar>
+      <header className={classes.header}>
+        <AppBar position="fixed" color="#ffffff">
```
pyphilia, line 94: Factor out white color.

src/config/cssStyles.js
```diff
@@ -0,0 +1,11 @@
+// header
+export const HEADER_SHADOW_EFFECT =
+  '0px 2px 4px -1px rgb(0 0 0 / 30%), 0px 4px 5px 0px rgb(0 0 0 / 20%), 0px 1px 10px 0px rgb(0 0 0 / 12%)';
+export const HEADER_LOGO_HEIGHT = 60;
+export const HEADER_HEIGHT = 100;
+
+// collectionCard
+export const COLLECTION_CARD_SHADOW_EFFECT =
+  '0px 2px 4px -1px rgb(0 0 0 / 30%), 0px 4px 5px 0px rgb(0 0 0 / 20%), 0px 1px 10px 0px rgb(0 0 0 / 12%)';
```
pyphilia, line 9: Both shadows are actually the same? I think design wise it makes sense to have the same shadow on all elements, so I think you can have one variable only for the shadow definition.

pyphilia (changes requested):
🦒

src/config/cssStyles.js
louisewang1: This two are the same, but actually some shadow effect are different. We can have a `Default shadow effect`, while the special ones can be defined seperately.

PR #141, opened 2022-04-13
fix: update the downloadItem mutation call

### LINKED ISSUE #140: Item Download mutation failed
When download an item, the endpoint url does not contain the correct item id.

louisewang1:
close #140


pyphilia (approved):
👍

PR #152, opened 2022-05-12
129/test and bug fix

### LINKED ISSUE #129: tests for new features

louisewang1:
close #129 
- fix the text-overflow on children cards
- tests for cc license
- tests for my lists


cypress/fixtures/items.js
```diff
@@ -254,3 +255,5 @@ export const GRAASPER_ITEMS = [
     categories: [{ categoryId: SAMPLE_CATEGORIES[1].id }],
   },
 ];
+
+export const getNumberOfOwnPublishedItems = (memberId) => PUBLISHED_ITEMS.filter(({path, creator}) => path.split('.').length === 1 && creator === memberId).length
```
pyphilia, line 259: `path.split('.').length===1` -> Wouldn't it be easier to test if it contains `.` ?

cypress/fixtures/members.js
```diff
@@ -65,3 +65,6 @@ export const buildMemberWithFavorites = (favoriteItems) => ({
   ...CURRENT_USER,
   extra: { ...CURRENT_USER.extra, favoriteItems },
 });
+
+// directly use id instead import from item fixtures here to avoid loop in export
+export const FAVORITE_ITEMS = ['ecafbd2a-5688-11eb-ae93-0242ac130002', 'egafbd2a-5688-11eb-ae93-0242ac130002'];
```
pyphilia, line 70: If you have a loop in export, it means you should define your variable elsewhere. Why not in items directly? Or in a new file.

cypress/integration/collection/ccLicense.spec.js
```diff
@@ -0,0 +1,29 @@
...
+describe('CC License in Summary', () => {
+  buildPublicAndPrivateEnvironments().forEach((environment) => {
+    it(`Display item's cc license for ${environment.currentMember.name}`, () => {
+      cy.setUpApi(environment);
+
+      const item = PUBLISHED_ITEMS[0];
+      cy.visit(buildCollectionRoute(item.id));
+      cy.wait(COLLECTION_LOADING_TIME);
+
+      cy.get(`#${SUMMARY_CC_LICENSE_CONTAINER_ID}`).should('be.exist');
```
pyphilia, line 16: Would it be possible to verify if the CC license is correct? (between the 2 possible options)

cypress/integration/myLists/myFavorites.spec.js
```diff
@@ -0,0 +1,29 @@
...
+  MY_FAVORITES_COLLECTIONS_GRID,
+} from '../../support/selectors';
+import { ITEM_PUBLISHED_TAG } from '../../fixtures/itemTags';
+import { buildMemberWithFavorites, FAVORITE_ITEMS } from '../../fixtures/members';
+
+describe('My Favorite Items', () => {
+  describe('Current user', () => {
+    // check if title and headings are displayed correctly
+    it('display favorite items', () => {
+      cy.setUpApi({currentMember: buildMemberWithFavorites(FAVORITE_ITEMS), items: PUBLISHED_ITEMS});
```
pyphilia, line 13: Instead of creating by hand `FAVORITE_ITEMS`, why not pick them from `PUBLISHED_ITEMS` before `setUpApi`?

cypress/integration/myLists/myFavorites.spec.js
```diff
@@ -0,0 +1,29 @@
...
+      cy.wait('@getCurrentMember');
+      cy.get(MY_FAVORITES_COLLECTIONS_GRID)
+      .children()
+      .should(
+        'have.length',
+        FAVORITE_ITEMS.length,
+      );
+    });
+  });
+});
```
pyphilia, line 29: Could we test that the application doesn't crash if there's a non-existent favorite item in the user's list?

cypress/integration/myLists/myLikes.spec.js
```diff
@@ -0,0 +1,34 @@
...
+    it('display liked items', () => {
+      cy.setUpApi({currentMember: CURRENT_USER, items: PUBLISHED_ITEMS});
+      cy.visit(MY_LIST_ROUTE);
+
+      // click my likes tab
+      cy.get(`#${buildMyListNavigationTabId(MY_LIST_TABS.MY_LIKES)}`).click();
+
+      cy.wait('@getPublicItemsWithTags').then(({ request: { url } }) => {
+        expect(url).to.contain(ITEM_PUBLISHED_TAG.id);
+      });
```
pyphilia, line 24: Why do you test the published tag for liked items?

cypress/integration/myLists/myFavorites.spec.js
pyphilia: Could we also test that this tab/section is not available if you are logged out?

cypress/integration/myLists/myLikes.spec.js
```diff
@@ -0,0 +1,34 @@
...
+      cy.wait('@getLikedItems');
+      cy.get(MY_LIKED_COLLECTIONS_GRID)
+      .children()
+      .should(
+        'have.length',
+        LIKED_ITEM_NUMBER,
+      );
+    });
+  });
+});
```
pyphilia, line 34: Could we also test that this tab/section is not available if you are logged out?

cypress/integration/myLists/myPublishments.spec.js
```diff
@@ -0,0 +1,33 @@
...
+import { MY_LIST_TABS } from '../../support/constants';
+
+describe('My Published Items', () => {
+  describe('Current user', () => {
+    // check if title and headings are displayed correctly
+    it('display published items', () => {
+      cy.setUpApi({currentMember: CURRENT_USER, items: PUBLISHED_ITEMS});
+      cy.visit(MY_LIST_ROUTE);
+
+      // click my likes tab
```
pyphilia, line 18: Remove?

cypress/support/constants.js
```diff
@@ -22,3 +22,10 @@ export const THUMBNAIL_EXTENSION = 'image/jpeg';
 export const HOME_PAGE_LOADING_TIME = 3000;
 export const ALL_COLLECTIONS_LOADING_TIME = 3000;
 export const COLLECTION_LOADING_TIME = 3000;
+
+// my list tabs
+export const MY_LIST_TABS = {
```
pyphilia, line 27: What does this do? If it's for selecting, you should apply "hardcoded" ids to the elements. Most probably you don't need a whole object just for testing.

src/config/selectors.js
```diff
@@ -18,3 +18,6 @@ export const GRAASP_SELECTION_TITLE_ID = 'graaspSelectionTitleId';
 export const DISCOVER_SECTION_TITLE_ID = 'discoverSectionTitleId';
 
 export const buildMyListNavigationTabId = (index) => `nav-tab-${index}`;
+
+export const MY_FAVORITES_COLLECTIONS_ID = 'favoriteCollectionsGrid';
+export const MY_LIKES_COLLECTIONS_ID = 'likedCollectionsGrid';
```
pyphilia, line 23: It might be my mistake, but we shouldn't have multiple `selectors.js` file. Let's keep only the `src/config/selectors.js` one.

cypress/fixtures/itemLikes.js
```diff
@@ -0,0 +1,15 @@
...
+    itemId: PUBLISHED_ITEMS[0].id,
+    memberId: MEMBERS.ANNA.id,
+  },
+  {
+    itemId: PUBLISHED_ITEMS[1].id,
+    memberId: MEMBERS.ANNA.id,
+  }
+]
+
+export const LIKED_ITEM_NUMBER = 1;
```
pyphilia, line 15: Can't we dynamically compute this number? It's obviously related to the number of liked items you return.

pyphilia (changes requested):
Thanks for the changes, I've left some comments 🐇

cypress/integration/myLists/myFavorites.spec.js
louisewang1: > Could we also test that this tab/section is not available if you are logged out?

I see this one is already tested in the header, when user logged out, they won't see the entrance on header thus the whole page won't be visible

cypress/support/constants.js
louisewang1: It's the index of tabs. It's not for selecting directly, but to generate the element ids because the element id of the tab contains the index. Otherwise I'll pass in 0, 1, 2 but it's not so clear which represent which.

cypress/integration/collection/ccLicense.spec.js
louisewang1: The displayed element does not contain the value of the option, so it's hard to check that.

cypress/integration/myLists/myLikes.spec.js
louisewang1: The liked items also fetch the published items and then do a filtering

cypress/integration/myLists/myLikes.spec.js
louisewang1: Same as above

cypress/integration/myLists/myLikes.spec.js
pyphilia: Could you add this as a comment? We will probably have to change later the mechanism with a proper endpoint (GET /items/liked or similar).

PR #155, opened 2022-05-30
feat: add langauge category

### LINKED ISSUE #154: Extra category: language

louisewang1:
close #154

### DIFF
--- a/src/components/home/AllCollections.js
+++ b/src/components/home/AllCollections.js
@@ -145,10 +145,14 @@ function AllCollections() {
         ?.id,
     )
     ?.sort(compare);
+  const languageList = allCategories?.get(
+    categoryTypes?.find((type) => type.name === CATEGORY_TYPES.LANGUAGE)?.id,
+  );
 
   // state variable to record selected options
   const [selectedLevel, setSelectedLevel] = useState([]);
   const [selectedDiscipline, setSelectedDiscipline] = useState([]);
+  const [selectedLanguage, setSelectedLangauge] = useState([]);
 
   // state variable to control the side menu
   const [sideBarStatus, setSideBarStatus] = useState(true);
@@ -171,6 +175,10 @@ function AllCollections() {
       }
       case CATEGORY_TYPES.DISCIPLINE: {
         setSelectedDiscipline([]);
+        break;
+      }
+      case CATEGORY_TYPES.LANGUAGE: {
+        setSelectedLangauge([]);
         break;
       }
       default: {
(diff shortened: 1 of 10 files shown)


src/components/home/AllCollections.js
```diff
@@ -312,6 +324,40 @@ function AllCollections() {
...
+        {isCategoriesLoading ? (
+          <Skeleton height="10%" />
+        ) : (
+          <List dense className={classes.list}>
+            {languageList?.map((entry) => (
+              <ListItem
+                button
+                key={entry.id}
+                onClick={handleClickForLanguage(entry.id)}
+                selected={selectedLanguage.indexOf(entry.id) !== -1}
```
pyphilia, line 345: I think you can use `.find` instead 
```
selectedLanguage.find(l => l.id === entry.id)
```

src/components/home/AllCollections.js
```diff
@@ -145,10 +145,14 @@ function AllCollections() {
...
     )
     ?.sort(compare);
+  const languageList = allCategories?.get(
+    categoryTypes?.find((type) => type.name === CATEGORY_TYPES.LANGUAGE)?.id,
+  );
 
   // state variable to record selected options
   const [selectedLevel, setSelectedLevel] = useState([]);
   const [selectedDiscipline, setSelectedDiscipline] = useState([]);
+  const [selectedLanguage, setSelectedLangauge] = useState([]);
```
pyphilia, line 155: We can use plurals for these arrays since we can choose multiple choices, right?

src/components/home/AllCollections.js
```diff
@@ -365,6 +413,7 @@ function AllCollections() {
             <LevelCollectionsPage
               selectedLevel={selectedLevel}
               selectedDiscipline={selectedDiscipline}
+              selectedLanguage={selectedLanguage}
```
pyphilia, line 416: Use plural?

pyphilia (changes requested):
Looks nice! I've added some comments. Do you think it would be possible to create a component for displaying the categories? I think it's three times the same code currently.

louisewang1:
image

Split the language out of other categories, and use different color outline different category types. The secondary color is a bit to red, use other colors? @pyphilia

src/components/home/AllCollections.js
```diff
@@ -195,10 +192,17 @@ function AllCollections() {
...
+    selectedDisciplines,
+    setSelectedDisciplines,
+  );
+  const handleClickForLevel = buildHandleClick(
+    selectedLevels,
+    setSelectedLevels,
+  );
+  const handleClickForLanguage = buildHandleClick(
+    selectedLanguages,
+    setSelectedLangauges,
```
pyphilia, line 204: Typo `setSelectedLanguages` (a-u misplaced)

pyphilia (changes requested):
Looks good to me! There's a minor typo. Also, would it be possible to write some tests? (layout tests depending on if the item has a language or not).
The categories are nicely divided. But maybe let's not use red, let's use everything in blue and we'll see later 🎈

PR #158, opened 2022-06-13
feat: display co-editors

louisewang1:
#157


cypress/fixtures/items.js
```diff
@@ -26,6 +26,7 @@ export const PUBLISHED_ITEMS = [
     settings: {
       tags: ['one-tag', 'two-tag'],
       ccLicenseAdaption: 'allow',
+      displayCoEditors: 'yes',
```
pyphilia, line 29: Shouldn't it be a boolean?

pyphilia (changes requested):
Looks good to me, I've added a concern about the data type 🥥

PR #181, opened 2022-08-11
refactor: move i18n to client wrapper to avoid resetting data

### LINKED ISSUE #180: Switch between languages clear data

pyphilia:
This PR fixes the reset on language switch. It happened because the render happens at the `app` level, which resets some data. Moving the language handling to `Wrapper` solved it.

It might be interesting to enable language routing at some point.

close #180


.husky/pre-push
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn lint && yarn build
+# yarn lint && yarn build
```
spaenleh, line 4: Do we want to keep this commented-out ?

spaenleh (approved):
🕹 From my limited knowledge of Next.js i would say it looks good. 
This change means that the title of the window will never be translated ? Maybe this is not a big deal 🤷🏻‍♂️

dialexo (approved):

dialexo:
> This change means that the title of the window will never be translated ? Maybe this is not a big deal 🤷🏻‍♂️

I don't know if we'll want to translate the product name anyway. Though I think we can always add the following in the translation effect if needed:
```diff
  React.useEffect(() => {
    // change language
    const lang = getLangCookie();
    if (lang) {
      i18n.changeLanguage(lang);
+     document.title = ...
    }
  }, []);
```

PR #184, opened 2022-08-19
fix: graasp-library/src/config/constants.js Line 49 refactored

IamGuneet:
fix:  fix: graasp-library/src/config/constants.js Line 49 deleted and refactored.

### DIFF
--- a/src/config/constants.js
+++ b/src/config/constants.js
@@ -46,7 +46,7 @@ export const MAIL_BREAK_LINE = '%0D%0A';
 
 export const LEFT_MENU_WIDTH = 300;
 
-export const GRAASP_BUILDER_URL = 'https://builder.graasp.org';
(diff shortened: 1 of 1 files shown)


pyphilia:
Hey! Thanks for your contribution but that was not what I was thinking for a fix. Really sorry for not being clear^^'

For this issue we want to get rid of the hardcoded string. It is another client URL we want to refer to. It is dynamic given the environment variable `GRAASP_BUILDER_HOST` since we deploy this project in development. The goal is to delete the variable `GRAASP_BUILDER_URL` and reuse `GRAASP_BUILDER_HOST` in the other file(s) instead (or create an intermediate variable in `constants.js` with a default value). Let me know if it makes sense!

IamGuneet:
> to 
Oh okay, no issue , its above my current level ig, thank you for trying to explain .

pyphilia:
No problem @IamGuneet ! Thanks for passing by

PR #200, opened 2022-11-14
feat: add dynamic tree viewer for copy modal

### LINKED ISSUE #75: Use Tree Viewer component from graasp-ui
If we use the graasp-ui tree component, it will display all children (instead of only the first level as currently).

dialexo:
With the dynamic tree view, "my items" and "shared items" are expanded (first-level) by default

Closes #75

### DIFF
--- a/src/components/collection/TreeModal.js
+++ b/src/components/collection/TreeModal.js
@@ -1,127 +1,145 @@
-import PropTypes from 'prop-types';
+import dynamic from 'next/dynamic';
+import PropTypes, { arrayOf } from 'prop-types';
 
 import React, { useContext, useState } from 'react';
-import { withTranslation } from 'react-i18next';
+import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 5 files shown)


src/components/collection/TreeModal.js
```diff
@@ -1,127 +1,158 @@
...
+    switch (prevent) {
+      case TREE_PREVENT_SELECTION.SELF_AND_CHILDREN:
+        // if the previous item is disabled, its children will be disabled
+        // and prevent selection on self
+        return parentIsDisabled || itemIds.find((x) => x === iId);
+      case TREE_PREVENT_SELECTION.NONE:
+      default:
+        return false;
+    }
+  };
```
pyphilia, line 69: I think you can remove this, since you will only copy.

pyphilia (approved):
LGTM, I've just added a minor comment 🔉

PR #211, opened 2022-12-13
chore: add release-please

spaenleh:
This PR adds release-please to manage app versions with PRs and create releases in GitHub.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,48 @@
+# Automate releases of new app versions
+name: release-please
+
+on:
+  push:
+    branches:
+      - "master"
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          release-type: node
+          package-name: graasp-library
+          changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"docs","section":"Documentation","hidden":false},{"type":"test","section":"Tests","hidden":false}]'
+
+      - uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      # Trigger an 'on: repository_dispatch' workflow to run in graasp-deploy repository
+      - name: Push tag to Graasp Deploy (Staging)
+        if: ${{ steps.release.outputs.release_created }}
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.REPO_ACCESS_TOKEN }}
+          repository: graasp/graasp-deploy
+          event-type: update-staging-version
+          client-payload: ${{steps.set-tag.outputs.json}}
```
dialexo, line 48: This only updates the corresponding version tag in graasp-deploy right? It does not actually trigger the staging deployment (as intended), is that correct? By the way, what happens if this is triggered twice (by 2 different newer version tags), but the stack is not released to staging for some reason? I assume only the later tag is considered?

dialexo (approved):
Just a quick question to clear any confusion 💡

.github/workflows/release-please.yml
spaenleh: Yes, this step only dispatches the new version to the graasp-deploy repo to let it know that a new version is available.
This actually creates a new file with the tag inside it. If you run this twice (a new version is created before the previous one is pushed to stage/prod), then only the newest written file will be considered. 

One caveat is that this step only runs if a release is created, so if for some reason the last step fails, re-running the workflow will not re-trigger the dispatch as the release will not be created. This has very little chance of happening and if it would, just pushing an empty commit with:
```
chore: release <your version>

Release-As: <your version>
```
which will trigger a new release.

I hope this helps 💡 !

PR #212, opened 2023-01-23
feat: use member's lang when signed in

### LINKED ISSUE #208: Switch language when user is logged in
When user is logged in, the interface should change to the language set in the user's settings.

Currently the language detection is only performed at the cookie level:
https://github.com/graasp/graasp-library/blob/51f49f80d41ac20361e703ce47b25c86cec35f64/src/components/common/Wrapper.js#L53-L59

We should add a request when the user is logged in

pyphilia:
close #208

### DIFF
--- a/src/components/common/TranslationWrapper.tsx
+++ b/src/components/common/TranslationWrapper.tsx
@@ -1,0 +1,28 @@
+import React, { useContext } from 'react';
+import { I18nextProvider } from 'react-i18next';
+
+import { getLangCookie } from '@graasp/sdk';
+
+import i18n from '../../config/i18n';
+import { QueryClientContext } from '../QueryClientContext';
+
+type Props = {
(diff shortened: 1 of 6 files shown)


src/components/common/TranslationWrapper.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+type Props = {
+  children: JSX.Element;
+};
+
+const TranslationWrapper = ({ children }: Props): JSX.Element => {
+  const { hooks } = useContext(QueryClientContext);
+  const { data: member } = hooks.useCurrentMember();
+
+  React.useEffect(() => {
```
spaenleh, line 17: Nitpick
```suggestion
  useEffect(() => {
```

spaenleh (approved):

dialexo (approved):

PR #215, opened 2023-01-26
fix: apply fixed colors to category chips

### LINKED ISSUE #214: Display discipline chips

pyphilia:
close #214


src/config/constants.ts
```diff
@@ -120,3 +120,9 @@ export const ENV = {
 };
 
 export const TREE_VIEW_MAX_WIDTH = 400;
+
+export const CATEGORY_COLORS = {
+  [CATEGORY_TYPES.DISCIPLINE]: '#bbdefb',
+  [CATEGORY_TYPES.LANGUAGE]: 'grey',
+  [CATEGORY_TYPES.LEVEL]: '#5050d2',
+};
```
dialexo, line 128: Does Typescript automatically infer that the type is `Record<CATEGORY_TYPE, string>` or is it `Record<string, string>`? E.g. is `CATEGORY_COLORS["foobar"]` allowed?

We may want to explicitly type it the former way, to ensure that the keyed access is always type-checked 😊

dialexo (approved):
LGTM but can you verify my comment? If it proves false, consider adding the change 🐯

PR #219, opened 2023-01-30
218 fix ga vars

### LINKED ISSUE #218: Fix GA vars
Rename Google analytics env variables to fix analytics not sent.

spaenleh:
This PR fixes some issues with the env variables.

- Domain var was not present in config
- Sentry var was not present in config
- GA var was renamed and wrong value was present in config
- Change org variables for app domain to use up-to-date ones `REACT_APP_DOMAIN_***`

closes #218


.github/workflows/cdelivery-ecs-library-caller.yml
```diff
@@ -28,9 +28,8 @@ jobs:
       ecs-cluster: ${{ secrets.ECS_CLUSTER_GRAASP_EXPLORE_STAGE }}
       ecs-service: ${{ secrets.ECS_SERVICE_GRAASP_EXPLORE_STAGE }}
       next-public-api-host: ${{ secrets.NEXT_PUBLIC_API_HOST_STAGE }}
-      next-public-domain: ${{ secrets.STAGE_GRAASP_DOMAIN }}
+      next-public-domain: ${{ secrets.REACT_APP_DOMAIN_STAGE }}
```
dialexo, line 31: Is it expected that the variables do not start with the usual `NEXT_` prefix in the secrets?

dialexo (approved):
I'm a bit confused by the naming but LGTM otherwise

.github/workflows/cdelivery-ecs-library-caller.yml
spaenleh: Yes this is because we are using Org-level secrets that refer to `graasp.org` or `dev.graasp.org` etc ... so they were created with React projects in mind, this is why they do not really have the naming convention...

spaenleh: And this is only on our CI/CD pipeline side, the local variables are all `NEXT_` prefixed as usual.

PR #224, opened 2023-02-13
feat: add platform switch to library

### LINKED ISSUE #222: Add platform switch
Integrate https://github.com/graasp/graasp-ui/issues/215 into Library

dialexo:
Closes #222 

Q: should we harmonize the navigation between the platforms at some point? Currently in Library, the spacing on the left of the logo is not the same as the other platforms (since the sidebar controls are absent), and adding artificial margins would look weird

[Kapture 2023-02-13 at 12.09.27.webm](https://user-images.githubusercontent.com/14943421/218443335-0f6aa026-6d28-43fb-9ed1-30f4736af80b.webm)


src/components/common/HeaderNavigation.tsx
```diff
@@ -0,0 +1,166 @@
...
+  color: 'inherit',
+  display: 'flex',
+  alignItems: 'center',
+};
+
+const StyledLink = styled(Link)(() => linkStyles);
+
+interface HeaderNavigationProps {
+  rootId?: string;
+  topItemName?: string;
```
spaenleh, line 116: What do you want to use this prop for ? It does not seem to get a value where this component is called.

spaenleh (approved):
🚢 LGTM ! Thank you for this, it looks super cool !

src/components/common/HeaderNavigation.tsx
dialexo: Good catch! Oops it was a copy paste from the component in Builder, but the prop is not used in LIbrary  👀

src/components/collection/Collection.js
```diff
@@ -115,26 +118,32 @@ const Collection = ({ id }) => {
...
-          </Alert>
-        )}
+        {
+          // show alert only if 1. user is logged in, 2. it has at least read access and 3. item is not published
+          currentMember?.id && canRead && !isPublished && (
+            <Alert severity="warning">
+              You are viewing this item in Library preview mode. It cannot be
+              viewed publicly.
+              {
+                // if the user is the admin of the item, also suggest publishing from Buider
```
spaenleh, line 128: Nitpick:
```suggestion
                // if the user is the admin of the item, also suggest publishing from Builder
```

spaenleh (approved):
🦅 LGTM! Thank you for the update !

PR #264, opened 2023-05-11
fix: display children card actions below item

### LINKED ISSUE #263: Can not copy from children cards
When clicking on the childrenCard copy button, the modal opens and when you select a destination, the page navigates to the children. This is because of the onClick event on the card that is still propagated even if it is explicitly stopped...

dialexo:
Closes #263 

image


spaenleh (approved):
💪

PR #268, opened 2023-05-11
fix: get publication settings from item itself and not its top-level parent

dialexo:
This behaviour is incorrect because the topmost parent may not be the published root.
Ideally we would want to fetch the published root, but this call is currently authenticated.
So meanwhile, we display the item publication settings of the current item, such that changes on published items which are not themselves top-level roots (i.e in my items) are still visible


spaenleh (approved):
Looks good to me ! Thank you 📎

PR #325, opened 2023-07-25
fix: add instant search and update workflow to include show-notifications env

spaenleh:


src/components/filters/FilterHeader.tsx
```diff
@@ -255,6 +260,14 @@ const FilterHeader: FC<FilterHeaderProps> = ({
   //   },
   // ]);
 
+  useEffect(() => {
+    // eslint-disable-next-line no-console
+    console.log(categoryPreset);
```
dialexo, line 265: Remove?

src/components/pages/AllCollections.tsx
```diff
@@ -41,9 +42,17 @@ const AllCollections: React.FC<AllCollectionsProps> = () => {
 
   useEffect(() => {
     const { query } = router;
-    if (query && query[UrlSearch.KeywordSearch]) {
-      const keywordSearch = query[UrlSearch.KeywordSearch] as string;
-      setSearchKeywords(keywordSearch);
+    if (query) {
+      if (query[UrlSearch.KeywordSearch]) {
+        const keywordSearch = query[UrlSearch.KeywordSearch] as string;
```
dialexo, line 47: Info: why is the cast required here?

src/config/constants.ts
```diff
@@ -71,6 +64,7 @@ export const DEFAULT_THUMBNAIL_SIZE = ThumbnailSize.Medium;
 
 export const UrlSearch = {
   KeywordSearch: 's',
+  CategorySearch: 'cat',
```
dialexo, line 67: I think leaving `category` would be fine. Should we add Google Analytics' `_gl` param here?

dialexo (approved):
LGTM :fondue: I've left a few suggestions

src/components/pages/AllCollections.tsx
spaenleh: Because the return is a `string | string[]` for example when you specify an array of values with the repeat format. In my case I should get a string. I could check for it though.

PR #349, opened 2023-08-14
fix: only show like button to logged-in users

### LINKED ISSUE #347: Only show like button when user is logged in

spaenleh:
In this PR we fix:
- only show like button to logged in users

Logged out users will not see the like button.

closes #347


pyphilia (approved):
👍

pyphilia (approved):

PR #399, opened 2023-10-09
feat: add like button for unlogged users and redirect to login

### LINKED ISSUE #398: Show like button but gray it out if user is not logged in
A user can like a collection in the library, but the button is only shown when the user is logged in. We should display the button but disable it when the user is not logged in and tell them that they need to log in if they want to like it.

LinaYahya:
Screenshot from 2023-10-11 13-14-05

closes #398


src/components/collection/summary/SummaryHeader.tsx
```diff
@@ -66,22 +75,44 @@ const SummaryHeader: React.FC<SummaryHeaderProps> = ({
...
   const handleUnlike = () => {
-    if (!collection?.id || !member?.id) {
+    if (!collection?.id) {
       console.error('unable to unlike an item which id is undefined');
       return;
     }
+    if (!member?.id) {
+      openLoginSnackbarMessage();
+      return;
+    }
```
spaenleh, line 115: In my opinion it is very unlikely that a user could unlike an item if they are not logged in. But we can keep it like this.

src/langs/en.json
```diff
@@ -116,5 +116,6 @@
   "SUGGESTION_TO_ENABLE_IN_DEPTH_SEARCH_TEXT": "Cannot find what you looked for?<br/>Try to enable in-depth search to include content",
   "SEARCH_NO_RESULTS": "Nothing corresponds to your search",
   "SEARCH_RESULTS_LOAD_MORE": "Load more",
-  "CONTENT_CHIP": "Content"
+  "CONTENT_CHIP": "Content",
+  "SIGNIN_MESSAGE": "You need to login First"
```
spaenleh, line 120:
```suggestion
  "SIGNIN_REQUIRED_MESSAGE": "You need to sign in first"
```

spaenleh (changes requested):
Thank you for this PR !

Can you add a screenshot for reference ?
We are trying to use "sign in" everywhere instead of login. 

I am not sure if the Snackbar is visible enough (users might not look at the bottom of the page) ? Lets see. We can always change it later if it is an issue.

spaenleh (approved):
Awesome !

PR #401, opened 2023-10-09
feat: add liked items page

### LINKED ISSUE #400: Display liked collections of a user in the library
A user likes collections in the library, they may want to brows their liked collections. A page should show them those collections.

LinaYahya:
Screenshot 2023-10-27 at 09 02 54

closes #400


spaenleh (changes requested):
Nice to see this page make a comeback.

Mentioning @fbierlai since we talked about this some time back and he could give some ideas on how to improve this page from the perspective of a user.

A few general comments and suggestions:
- the route could be changed to `/liked` 
- the page should display a message if the user is not logged in
- we should not display the favorites in there in order to not confuse users, these are not published yet, so they should not appear in the library
- not sure if we keep the "published items" as we already have a similar menu in the builder and there is/will soon be a member page to see these same collections. 
- the container looks a bit wrong on some of the collections, check that they are centered and don't have weird widths
- i think we could also change the naming to "Liked Collections" instead of "My likes"

src/components/layout/UserSwitchWrapper.tsx
```diff
@@ -24,24 +28,35 @@ const UserSwitchWrapper: FC<Props> = ({ ButtonContent }) => {
...
-        )}
-      />
+    <Box sx={{ display: 'flex', gap: '24px', alignItems: 'center' }}>
+      {member?.id && (
+        <Link
+          href={MY_LIST_ROUTE}
+          style={{ textDecoration: 'none', color: 'white' }}
+        >
+          {t(LIBRARY.LIKED_ITEMS)}
+        </Link>
```
spaenleh, line 41: Can you use the same style as the button "All Collections" ? with the background effect on hover and the font size and weight ? this would make the header more uniform.

LinaYahya: @spaenleh 
Thanks for your comments, changes are implemented, But I kept the published items tab (check screenshot within PR description), Any recommendations?

src/components/pages/MyList.tsx
```diff
@@ -28,13 +29,33 @@ const { Main } = {
...
+        context={Context.Library}
+        headerLeftContent={leftContent}
+        headerRightContent={rightContent}
+      >
+        <Box p={5}>
+          <Error code={ERROR_UNAUTHORIZED_CODE} />
+        </Box>
+      </Main>
+    );
+  }
```
spaenleh, line 58: I know it has been implemented like this in a lot of our code base, but if we want to take full advantage of the `stale-while-revalidate` semantics that drive react-query, we should:
1. check if there is data and display it
2. check if there was an error
3. fallback to the loading state

This will ensure that we can fully exploit the fact that the stale data is kept in the cache while the refresh query is being sent out. For the use this means that we continue to display the "old" data until we get the new data, at which point it is silently replaced and updated in the UI without you needing to do anything. 
Specifically this will prevent seeing the loading indicators every time we refetch data.

In case you want to read more about it and get some practical examples, check out [this article from the maintainer of react-query](https://tkdodo.eu/blog/status-checks-in-react-query)

spaenleh (approved):
Thank you for the changes and your hard work. 
I left a comment to improve the behaviour of queries fetching. Feel free to try it. 

Before merging, I think we should have a quick demo with Denis to get his feedback.

spaenleh (approved):
Thank you for the changes, I took the liberty to adapte some of them, you will find my initial comments in the review.

PR #430, opened 2023-11-10
feat: upgrade to nextjs 14 and react 18

pyphilia:
The upgrade to nextjs 14 is not really easy, I tried to move to the app folder. 
The current status of the branch works, but is not connected to the query client from the server. I think there's a better way of handling server/vs client files.
Plus some minor css bugs.

- [ ] prefetch (for another issue)
- [ ]  backend connection
- [x] deployment env variables


src/config/env.ts
```diff
@@ -1,39 +1,27 @@
...
-  publicRuntimeConfig.NEXT_PUBLIC_GRAASPER_ID;
-export const S3_FILES_HOST = publicRuntimeConfig.NEXT_PUBLIC_S3_FILES_HOST;
+  process.env.NEXT_PUBLIC_GRAASP_ACCOUNT_HOST || 'http://localhost:3114';
+// eslint-disable-next-line prefer-destructuring
+export const NODE_ENV = process.env.NODE_ENV;
+export const GA_MEASUREMENT_ID = process.env.NEXT_PUBLIC_GA_MEASUREMENT_ID;
+export const GRAASPER_ID = process.env.NEXT_PUBLIC_GRAASPER_ID;
 export const SHOW_NOTIFICATIONS =
-  publicRuntimeConfig.NEXT_PUBLIC_SHOW_NOTIFICATIONS ||
+  process.env.NEXT_PUBLIC_SHOW_NOTIFICATIONS ||
```
spaenleh, line 19: Careful here because if the value is `"false"` bad things will happen. I think we can remove the first part.

src/config/queryClient.ts
```diff
@@ -3,7 +3,7 @@ import notifier from './notifier';
 
 // eslint-disable-next-line import/prefer-default-export
 export const QUERY_CLIENT_OPTIONS = {
-  API_HOST: GRAASP_API_HOST,
+  API_HOST: GRAASP_API_HOST!,
```
spaenleh, line 6: Provide a default in 'env.ts' to not have to use `!` operator.

spaenleh (approved):
Looks good, just 2 small comments, you can merge once done !

Issue #466, opened 2024-01-08
In collection view show rendered description on "show less"


pyphilia:
![Screenshot 2024-01-08 at 09 56 52](https://github.com/graasp/graasp-library/assets/11229627/67469182-1777-4436-9c64-bfa6feb9d1e2)
<img width="497" alt="Screenshot 2024-01-08 at 09 57 23" src="https://github.com/graasp/graasp-library/assets/11229627/8ec15a81-3ad5-4cf2-b3fb-ad16223c729d">

Solution: display rendered description with a fixed height and a fade effect.

PR #486, opened 2024-01-11
fix: control show less/more button at public profile bio

### LINKED ISSUE #480: Public Member Page: do not show "show more show less" if there's no description
Screenshot 2024-01-10 at 15 12 37
Screenshot 2024-01-10 at 15 12 51

LinaYahya:
closes #480

### DIFF
--- a/src/components/common/ShowLessAndMoreContent.tsx
+++ b/src/components/common/ShowLessAndMoreContent.tsx
@@ -21,7 +21,7 @@ const ShowLessAndMoreContent = ({
 
   const [isContentCollapsed, setIsContentCollapsed] = useState(true);
 
-  const handleShowMoreButton = () => {
+  const handleToggleButton = () => {
     setIsContentCollapsed(!isContentCollapsed);
   };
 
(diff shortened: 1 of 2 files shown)


pyphilia:
I'm not really convinced by the line-height-ref method. Plus there're some mismatching problem here:
- The member profile does not show html normally, but you use the content description that uses interweave that might render html. But your implementation handles text lines only.
- on the other side we have also a pb with content description because it does not show html if not expanded https://github.com/graasp/graasp-library/issues/466 

I think a low-level component should include the button that expand button + content with fading (mainly the issue I've linked). Maybe it should be adaptable to use or not interweave. What do you think @spaenleh ? So this PR would handle both issues.

spaenleh:
@pyphilia @LinaYahya I agree. Also if the computation actually just returns the line height, why not simply use the `lh` css unit ? 
I think component wise we should have:
- a `DescriptionWrapper` that receives a child (the description) which is rendered with a fixed height (2 or 3 `lh`) when showing less and un-constrained height when showing more + the button to do so.
- one of more Description components (one with interwaeave for html descriptions, one simple text for the bio description)

Then you compose them together to get the desired result

spaenleh (approved):

spaenleh (approved):

PR #503, opened 2024-01-18
fix: category chip translations, date locales for logged out user and published root fetch info for categories

spaenleh:


pyphilia (approved):
LGTM 🚋

PR #515, opened 2024-02-09
fix: new version of UI and no more dynamic imports

spaenleh:


src/components/pages/Home.tsx
```diff
@@ -1,6 +1,6 @@
 'use client';
 
-import dynamic from 'next/dynamic';
+// ^ the use client directive should be removed as soon as possible once query-client provides a real ESm build
```
ReidyT, line 3: Is there an issue open for that ?

ReidyT (approved):
Nice PR ! All looks good for me 🙌

src/components/pages/Home.tsx
spaenleh: Yes see https://github.com/graasp/graasp-query-client/issues/626

PR #527, opened 2024-02-26
fix: update esm dependencies

spaenleh:


cypress/e2e/allCollections/allCollections.cy.ts
```diff
@@ -18,6 +18,8 @@ import { buildPublicAndPrivateEnvironments } from '../../fixtures/environment';
 import { PUBLISHED_ITEMS } from '../../fixtures/items';
 import { getRootPublishedItems } from '../../support/utils';
 
+const i18n = i18nConfig();
```
pyphilia, line 21: Was it important to have a function here?

src/components/common/TranslationWrapper.tsx
```diff
@@ -14,6 +14,8 @@ const TranslationWrapper = ({ children }: Props): JSX.Element => {
   const { hooks } = useContext(QueryClientContext);
   const { data: member } = hooks.useCurrentMember();
 
+  const [i18n] = useState(i18nConfig);
```
pyphilia, line 17: ?
```suggestion
  const [i18n] = useState(i18nConfig());
```

src/config/notifier.tsx
```diff
@@ -8,9 +8,10 @@ import { FAILURE_MESSAGES, SUCCESS_MESSAGES } from '@graasp/translations';
 import ToastrWithLink from '../components/common/ToastrWithLink';
 import LIBRARY from '../langs/constants';
 import { SHOW_NOTIFICATIONS } from './env';
-import i18n from './i18n';
+import { i18nConfig } from './i18n';
 import { buildPlayerViewItemRoute } from './paths';
 
+const i18n = i18nConfig();
```
pyphilia, line 14: Do you still keep the same language since you have a different config?

pyphilia (approved):
I have some concerns about your i18n config, but I trust you if it works.

src/components/common/TranslationWrapper.tsx
spaenleh: Actually here I use the functional form of the useState to create the instance for the lifetime of the component. it is similar to:
```ts
const. [i18n] = useState(() => i18nConfig());
```

src/config/notifier.tsx
spaenleh: Ah yes that might be an issue ... not sure how I should get the same instance from the `TranslationWrapper` ? D you have an idea ?

PR #689, opened 2024-09-17
feat(deps): update TanStack Query to v5

ReidyT:


spaenleh (approved):
Looks good to me !

I guess we can't yet update the version of the tanstack query package here ? Maybe in a later PR.

src/components/collection/UnpublishedItemAlert.tsx
```diff
@@ -15,7 +15,7 @@ type Props = {
   canRead: boolean;
   canPublish: boolean;
   isPublished: boolean;
-  currentMember?: CompleteMember | null;
+  currentMember?: CompleteMember | CompleteGuest | null;
```
pyphilia, line 18:
```suggestion
  currentMember?: CurrentAccount;
```

src/components/common/TranslationWrapper.tsx
```diff
@@ -10,13 +10,23 @@ type Props = {
   children: JSX.Element | JSX.Element[];
 };
 
+const getUserLang = (
+  member: CompleteGuest | CompleteMember | undefined | null,
```
pyphilia, line 14:
```suggestion
  member?: CurrentAccount,
```

pyphilia (approved):
Some improvements for the type, but LGTM! 🛕

src/components/collection/UnpublishedItemAlert.tsx
ReidyT: Nice thanks!

ReidyT: But as discussed with @spaenleh, I should return null if it is not an real account.

PR #699, opened 2024-10-03
feat: use prefetch in the home page

ReidyT:
- feat(deps): update the QueryClient dev deps to tanstack v5 (was v3)

For now, the prefetch never refetch the data, except when the `refetchOnMount` is set to true. If the prefetch code is commented, all works correctly (except that's not SSR)...


app/providers.tsx
```diff
@@ -53,7 +53,6 @@ export default function Providers(props: { children: React.ReactNode }) {
   return (
     <QueryClientProvider client={client}>
       {children}
-      {/* <ReactQueryDevtools initialIsOpen={false} /> */}
```
spaenleh, line 56: Can we enable the devtools with ssr ? I think it would be nice :)

package.json
```diff
@@ -33,7 +33,7 @@
     "@emotion/react": "11.13.0",
     "@emotion/server": "11.11.0",
     "@emotion/styled": "11.13.0",
-    "@graasp/query-client": "3.26.0",
+    "@graasp/query-client": "github:graasp/graasp-query-client#947-prefetch-home-page",
```
spaenleh, line 36: reminder

spaenleh (approved):
This looks really great ! Wonderfull work 💪 !

PR #701, opened 2024-10-07
fix: show empty accessible

### LINKED ISSUE #700: Copy modal show error alert when accessible is empty

pyphilia:
fix #700


ReidyT (approved):
LGTM 🪐

src/components/collection/copyModal/AccessibleNavigationTree.tsx
```diff
@@ -18,13 +19,15 @@ interface AccessibleNavigationTreeProps {
...
 const AccessibleNavigationTree = ({
+  isDisabled,
   onClick,
   onNavigate,
   selectedId,
 }: AccessibleNavigationTreeProps): JSX.Element => {
-  const { t } = useLibraryTranslation();
-  const [page, setPage] = useState(1);
   const { hooks } = useContext(QueryClientContext);
+  // todo: to change with real recent items (most used)
```
spaenleh, line 28: The comment is not at the right place no ?

src/components/collection/CollectionsGrid.tsx
```diff
@@ -40,8 +40,8 @@ const CollectionsGrid = ({
         justifyContent="flex-start"
         id={id}
       >
-        {Array.from({ length: 4 }).map(() => (
-          <Grid xs={6} sm={4} md={3} lg={3} xl={2}>
+        {Array.from({ length: 4 }, (_, idx) => idx).map((idx) => (
```
spaenleh, line 43: I think in cases like this, we should pre-generate a list with fixed ids and use that to map over. Using the index completely defies the purpose of using a key in react.

spaenleh (approved):

src/components/collection/CollectionsGrid.tsx
pyphilia: It's kinda the same, except I used index to fill the map.

PR #712, opened 2024-10-23
feat: use item lang instead of category lang

pyphilia:
- [x] update tests


src/components/collection/Collection.tsx
```diff
@@ -68,47 +64,50 @@ const Collection = ({ id }: Props) => {
...
-        <Error code={ERROR_UNEXPECTED_ERROR_CODE} />
-      </Box>
+      <>
+        <UnpublishedItemAlert
+          itemId={id}
+          canRead={canRead}
+          canPublish={canPublish}
+          isPublished={
+            isLoadingPublishedEntry ||
+            (!!itemPublishEntry && !isErrorPublishedEntry)
```
ReidyT, line 79: I know it's not you, but I would remove the double negation ^^

```suggestion
            (Boolean(itemPublishEntry) && !isErrorPublishedEntry)
```

src/components/collection/summary/SummaryDetails.tsx
```diff
@@ -126,10 +127,11 @@ const SummaryDetails: React.FC<SummaryDetailsProps> = ({
...
 
-  // TODO: should use item language
-  const languages = itemCategories
-    ?.filter((c) => c.category.type === CategoryType.Language)
-    ?.map((c) => c.category);
+  let langValue = langs[DEFAULT_LANG];
+  if (collection.lang in langs) {
+    // @ts-ignore
+    langValue = langs[collection.lang];
+  }
```
ReidyT, line 134: Why do you need `@ts-ignore`? Also what do you think about that?

```suggestion
  const lang = collection.lang in langs ? collection.lang : DEFAULT_LANG;
  const langValue = langs[lang];
```

src/components/filters/CategoryFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+  buildSearchFilterPopperButtonId,
+} from '../../config/selectors';
+import { Filter } from './Filter';
+
+type FilterProps = {
+  category: string;
+  title: string;
+  options?: Category[];
+  // IDs of selected options.
+  selectedOptions: string[];
```
ReidyT, line 17:
```suggestion
  selectedOptionIds: string[];
```

src/components/filters/Filter.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+import { GRAASP_COLOR } from '../../config/constants';
+import { useLibraryTranslation } from '../../config/i18n';
+import LIBRARY from '../../langs/constants';
+import { FilterPopper, FilterPopperProps } from './FilterPopper';
+
+type FilterProps = {
+  title: string;
+  // IDs of selected options.
+  selectedOptions: string[];
```
ReidyT, line 14:
```suggestion
  selectedOptionIds: string[];
```

src/components/filters/Filter.tsx
```diff
@@ -0,0 +1,137 @@
...
+  isLoading?: boolean;
+  onClearOptions: FilterPopperProps['onClearOptions'];
+  onOptionChange: FilterPopperProps['onOptionChange'];
+  id: string;
+  buttonId: string;
+  options?: FilterPopperProps['options'];
+};
+
+// eslint-disable-next-line import/prefer-default-export
```
ReidyT, line 23: should we export default and remove this?

src/components/filters/CategoryFilter.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onOptionChange: (key: string, newValue: boolean) => void;
+  onClearOptions: () => void;
+  isLoading: boolean;
+};
+
+// eslint-disable-next-line react/function-component-definition
```
ReidyT, line 23: same here? or should we remove this eslint rule?

src/components/filters/LangFilter.tsx
```diff
@@ -0,0 +1,45 @@
...
+import {
+  SEARCH_FILTER_LANG_ID,
+  SEARCH_FILTER_POPPER_LANG_ID,
+} from '../../config/selectors';
+import { Filter } from './Filter';
+
+type LangFilterProps = {
+  title: string;
+  // IDs of selected options.
+  selectedOptions: string[];
```
ReidyT, line 12:
```suggestion
  selectedOptionIds: string[];
```

ReidyT (approved):
Thanks for this PR, LGTM! I just let you some suggestions 🎃

src/components/filters/CategoryFilter.tsx
pyphilia: For now I will let it, otherwise I will need to change all other components. I will create an issue.

PR #732, opened 2025-01-06
feat: add storybook and collapsible text

pyphilia:
Add storybook in library, since we won't be using graasp-ui anymore. Instead we will get graasp-ui from /client.
Whenever possible, library components should stay in library, like the new card layout.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,0 +1,17 @@
+import type { StorybookConfig } from '@storybook/nextjs';
+
+const config: StorybookConfig = {
+  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
+  addons: [
+    '@storybook/addon-onboarding',
+    '@storybook/addon-essentials',
+    '@chromatic-com/storybook',
+    '@storybook/addon-interactions',
+  ],
+  framework: {
+    name: '@storybook/nextjs',
+    options: {},
+  },
+  staticDirs: ['../public'],
+};
+export default config;
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -2,12 +2,7 @@ import type { StorybookConfig } from '@storybook/nextjs';
 
 const config: StorybookConfig = {
   stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
-  addons: [
-    '@storybook/addon-onboarding',
-    '@storybook/addon-essentials',
-    '@chromatic-com/storybook',
-    '@storybook/addon-interactions',
-  ],
+  addons: ['@storybook/addon-essentials', '@storybook/addon-interactions'],
   framework: {
     name: '@storybook/nextjs',
     options: {},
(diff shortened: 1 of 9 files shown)


.storybook/main.ts
```diff
@@ -0,0 +1,17 @@
+import type { StorybookConfig } from '@storybook/nextjs';
+
+const config: StorybookConfig = {
+  stories: ['../src/**/*.mdx', '../src/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
+  addons: [
+    '@storybook/addon-onboarding',
```
spaenleh, line 6: You can remove this one.

.storybook/main.ts
```diff
@@ -0,0 +1,17 @@
...
+    '@storybook/addon-essentials',
+    '@chromatic-com/storybook',
```
spaenleh, line 8: Also this one since we are not using their hosted service.

package.json
```diff
@@ -71,9 +73,17 @@
     "uuid": "10.0.0"
   },
   "devDependencies": {
+    "@chromatic-com/storybook": "^3.2.3",
```
spaenleh, line 76: remove

package.json
```diff
@@ -71,9 +73,17 @@
...
     "@commitlint/cli": "19.5.0",
     "@commitlint/config-conventional": "19.5.0",
     "@cypress/code-coverage": "3.13.4",
+    "@storybook/addon-essentials": "^8.4.7",
+    "@storybook/addon-interactions": "^8.4.7",
+    "@storybook/addon-onboarding": "^8.4.7",
```
spaenleh, line 82: remove

src/components/common/CollapsibleText/CollapsibleText.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+  component: CollapsibleText,
+} satisfies Meta<typeof CollapsibleText>;
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+export const Collapsed = {
+  args: {
+    content:
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus scelerisque, nibh eu dapibus dignissim, eros orci tristique ipsum, id interdum ex neque a dui. Quisque pretium aliquam dui, pulvinar venenatis sem consectetur non. Sed felis mi, viverra sit amet blandit sed, vestibulum eu purus. Nulla eget tellus sodales, volutpat nulla in, blandit tellus. Fusce congue vitae elit ac scelerisque. Maecenas fringilla ipsum in enim volutpat dignissim. Aenean convallis urna vel nibh semper faucibus. Nunc non pellentesque nibh. Sed tempor, erat sit amet volutpat placerat, metus quam auctor nunc, in consectetur orci dui quis ligula.',
```
spaenleh, line 17: You can use the `args` in the meta for default arg values. For `content` and `numberOfLinesToShow` I think it would make sense since you are not changing them in both stories.

spaenleh (changes requested):
This looks good. 
I suggest removing some storybook addons we will not need and using meta default args for the props that will not change in your stories.

Otherwise great work ! 🍣

spaenleh (approved):
🎉

PR #739, opened 2025-01-28
refactor: use tanstack query and remove react query

pyphilia:
Remove `react-query` to use `@tanstack/query`


package.json
```diff
@@ -44,6 +44,7 @@
     "@mui/lab": "6.0.0-beta.13",
     "@mui/material": "6.1.5",
     "@sentry/nextjs": "8.35.0",
+    "@tanstack/react-query": "5.65.1",
```
spaenleh, line 47: Do you also want to add the devtools ?

spaenleh (approved):
Nice !

package.json
pyphilia: Not particularly, but it can be added later if necessary.

PR #740, opened 2025-01-28
fix: update home endpoints with generated api

pyphilia:
Update broken endpoints (liked, recent, get by member) using auto generated api.

Includes generated client. This PR does not include a process to update the client files.

based on https://github.com/graasp/graasp-library/pull/739

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -34,7 +34,7 @@ export const ENV = {
 
 export const TREE_VIEW_MAX_WIDTH = 400;
 
-export const HOMEPAGE_NB_ELEMENTS_TO_SHOW = 8;
+export const HOMEPAGE_NB_ELEMENTS_TO_SHOW = 6;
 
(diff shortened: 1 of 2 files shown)


src/config/constants.ts
```diff
@@ -34,7 +34,7 @@ export const ENV = {
 
 export const TREE_VIEW_MAX_WIDTH = 400;
 
-export const HOMEPAGE_NB_ELEMENTS_TO_SHOW = 12;
+export const HOMEPAGE_NB_ELEMENTS_TO_SHOW = 8;
```
spaenleh, line 37: The number of cards on the page has changed ? is it not a multiple of 2, 3 and 4 ?

spaenleh (approved):
Changes look good, I have not run the code, so I can't say how it works. But it looks good to see the new generation being used !!!

src/config/constants.ts
pyphilia: The column is 3 max, so I can set 6 I guess?

PR #741, opened 2025-01-28
build: generate client from server

pyphilia:
Fetch openapi json from server (based on env) and generates the client sdk. 

Moved generation and generated files to `src/openapi`.

Use `openapi.json` input from url (eg. `api.graasp.org/docs/json`) so it's also usable locally if your local server is up.


spaenleh (commented):
Just so I understand, with this you can refresh the configuration by simply running `yarn openapi-ts` and it will fetch the json spec from the running server (using the env var) and generate the output. 

What is stored in version control is only this generated artifacts. 

My only comment would be to rename the package.json script, as the current name is not really easy to remember and to type.

spaenleh (approved):

PR #772, opened 2025-05-15
feat: refactor to tanstack start

spaenleh:


pyphilia:
General feedback: looks generally good! 

I notice a blank page, but it's not vital (slow 3g)

https://github.com/user-attachments/assets/53c950f1-1ff3-4785-9600-fdcb53651629

- enable in-depth search does not work
- clear search query does not work
- broken spanish translation: "Buscar entre 2 } recursos"
- current user thumbnail (top bar) does not work 

Not vital: would be nice to have the same menu as in client
Screenshot 2025-05-23 at 10 19 56

PR #800, opened 2025-07-07
fix: add active link styling on header

### LINKED ISSUE #790: Use active link style in navigation bar

spaenleh:
In this PR:

- fix #790

https://github.com/user-attachments/assets/5d3a3c56-567f-4b73-9647-6434feb72f71

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite dev --port 3002",
-    "start": "vite start",
+    "start": "pnpm dev",
     "build": "vite build",
     "build:test": "vite build --mode test",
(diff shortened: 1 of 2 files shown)


pyphilia:
Would it be easy to have a bottom border instead of an underline ? I think it would look better.
On another note, I think it would look good with the navigation being all bold, so it does not look like random excluded texts in some header.
 
Screenshot 2025-07-07 at 14 26 56

pyphilia:
Can't we display the border bottom instead of the link underline on hover? Not so fan of the double underline.

spaenleh:
@pyphilia I made the changes you suggested. Can you give your final review ?

PR #808, opened 2025-07-17
fix: use sort instead of toSorted for compatibility

### LINKED ISSUE #807: [COMPATIBILITY] `.toSorted()` is not available on older devices/browsers
We have received reports that [`.toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) is not available on some older devices or browsers.

- Android Chrome with version lower than 110 (2023)
- Mobile Safari on iOS with version lower than 16 (2022)

In our current usage we can revert to using [`.sort()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) which has a wider support since it was introduced before `.toSorted()`.

spaenleh:
- fix #807


pyphilia (approved):
🍭

PR #811, opened 2025-07-17
fix(ci): remove tests that are often flacky in safari

spaenleh:
In this PR I disable the flow > Search test in safari because it is very unstable in CI and often fails. 

I have tried to investigate why this happens, but with no luck. 
The safest path for us forward is to temporarily disable tests that are falcky so that we have a higher confidence in our test results.


pyphilia (approved):
😢

PR #826, opened 2025-08-26
feat: add copy to builder option

pyphilia:
Add copy option in collection page. This redirect to builder.


spaenleh (approved):

spaenleh:
@pyphilia The issue in the test was that the builder page had a query parameter in the URL, but we were specifying a full URL making the matching strict. So the intercept was not working. I opted for using a wildcard pattern for the intercept. It now works. 

I will merge so that we can deploy in the stack.

PR #833, opened 2025-09-16
feat: improve homepage

pyphilia:


src/config/constants.ts
```diff
@@ -19,9 +19,9 @@ export const MEMBER_AVATAR_MAIN_SIZE = 256;
 
 export const UrlSearch = {
   KeywordSearch: 's',
-  DisciplineTagSearch: 'discipline',
-  LevelTagSearch: 'level',
-  ResourceTypeTagSearch: 'resourceType',
+  DisciplineTagSearch: 'disciplines',
+  LevelTagSearch: 'levels',
+  ResourceTypeTagSearch: 'resourceTypes',
```
spaenleh: Why were these changed ? We need to check that it still works with the search params. I remember that there was a difference between the search params in the URL and the parameters that are passed to the backend.

spaenleh:
@pyphilia there was in issue in the Unit tests where the BigCard story was trying to access `process` and it was undefined.

This was actually due to the inclusion of `BigCardGrid` in `BigCard.tsx` which in turn was importing `CollectionItem` which was importing CollectionCard which was making a query for the thumbnail, thus importing a queryOptions and that was actually reading the heyApi config which imported `API_HOST` thus including process which did not exist.

I fixed it temporarily by moving BigCardGrid in its own file that is not tested by storybook.

The long term fix that I am working on currently is to include the thumbnails in the search results so we do not need to make another call to get them once we have a result. Since they are cached, I think it will be fine. These PRs will be ready soon.

src/components/common/Card/BigCardGrid.tsx
spaenleh: Not sure why the grid is always one element on top of the other (for featured)

spaenleh (approved):

PR #838, opened 2025-10-15
feat: use thumbnails from search

spaenleh:
In this PR:
- use thumnails from search results
- update the generated queries to the latest state of the backend
- remove unused deps from the excluded optimized deps in vitest

ref https://github.com/graasp/graasp/pull/2006


pyphilia (approved):
LGTM 🫥

PR #866, opened 2026-01-05
feat: update translations to use resources, search among more than

pyphilia:
Update translations files:
- remove some unused keys
- use "resources" as much as possible
- fix some translations in spanish
- fix search text by using "search among more than" to allow "search mong more than 1000 resources"


pyphilia:
ref https://github.com/graasp/graasp/issues/2038

spaenleh (approved):

